PR #5, opened 2024-09-17
Load Model Checkpoints with Arbitrary Output Dimensions

ramppdev:
This adds support for pretraining and subsequent finetuning on different datasets where the number of classes / the output dimension of the model differs.

If the output dimensions of the last layer found in the checkpoint and the current model differ, the weights of the last linear layer are skipped. This also applies to the optimizer state.
If both datasets (and therefore the state dict and model) share the same output dimension, `skip_last_layer: True` skips the last linear layer explicitly.

Previously, we supported a `pretrained` attribute in the  model, optimizer, and scheduler configurations.
These attributes are now all moved to the model configuration, as they should be grouped together, and are specified as follows:
- `model_checkpoint`: The path to the model checkpoint file.
  If the output dimensions between the checkpoint and the model configuration do not match,
  the checkpoint will be loaded without the last linear layer.
- `optimizer_checkpoint`: The path to the optimizer checkpoint file.
  If the output dimensions between the checkpoint and the model configuration do not match,
  the checkpoint will be loaded without the last linear layer, similar to the model checkpoint.
  If no model checkpoint is provided, :attr:`skip_last_layer` is automatically set to ``True``.
- `scheduler_checkpoint`: The path to the scheduler checkpoint file.

Also updated the docs explaining the usage of the new attributes (the old `pretrained` attribute was not mentioned) and added a simple integration test to cover loading all 3 checkpoints.

### DIFF
--- a/autrainer/training/utils.py
+++ b/autrainer/training/utils.py
@@ -1,0 +1,17 @@
+from typing import Dict, Optional
+
+
+def format_results(
+    results: Dict[str, float],
+    results_type: str,
+    training_type: str,
+    iteration: Optional[int] = None,
+) -> str:
+    s = f"{results_type} results"
+    s += f" at {training_type} {iteration}" if iteration else ""
+    s += ":\n"
+    max_key_len = max([len(k) for k in results.keys()])
+    s += "\n".join(
+        [f"{(k+':').ljust(max_key_len+1)} {v:.4f}" for k, v in results.items()]
+    )
+    return s
--- a/autrainer/training/utils.py
+++ b/autrainer/training/utils.py
@@ -1,4 +1,7 @@
+import os
 from typing import Dict, Optional
+
+import torch
 
 
 def format_results(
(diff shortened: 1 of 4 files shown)


autrainer/training/utils.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Returns:
+        True if the last linear layer was skipped, otherwise False.
+    """
+    last_linear_layer = None
+    for name, layer in model.named_modules():
+        if isinstance(layer, torch.nn.Linear):
+            last_linear_layer = name
+
+    if last_linear_layer is None:
```
millinma, line 47: I don't think we should raise an error if there is no linear layer? Could be a fully convolutinal NN with a final pooling layer and sigmoid for binary classification for instance that should be a viable option to load too?

docs/source/modules/models.rst
```diff
@@ -23,6 +23,19 @@ To avoid race conditions when using :ref:`hydra_launcher_plugins` that may run m
...
 
+To optionally use model, optimizer, or scheduler checkpoints, the following attributes can be set in any model configuration file:
+
+* :attr:`model_checkpoint`: The path to the model checkpoint file.
+  If the output dimensions between the checkpoint and the model configuration do not match,
+  the checkpoint will be loaded without the last linear layer.
+* :attr:`optimizer_checkpoint`: The path to the optimizer checkpoint file.
+  If the output dimensions between the checkpoint and the model configuration do not match,
+  the checkpoint will be loaded without the last linear layer, similar to the model checkpoint.
+  If no model checkpoint is provided, :attr:`skip_last_layer` is automatically set to ``True``.
```
millinma, line 34: I'm wondering if default should be false, in case it's more of a continued training, maybe this is an equally likely use case as the change of dataset?

millinma (changes requested):

autrainer/training/utils.py
ramppdev: Good point. In that case loading would fail. Instead of throwing an error we could wrap `model.load_state_dict` in a try catch block and fail if we are unable to load the features.
However if there is no linear layer present, it is hard to find out what has to be removed from the state dict.
In theory we can also traverse the `named_modules` of the model along the weights and biases of the state dict and check for matching shapes, then keep only the weight up until the last matching shape.
What approach would you perfer?

docs/source/modules/models.rst
ramppdev: Continuing training is already implemented, if a run exists with the same configuration and only differs by the number of iterations (smaller), it is automatically continued (as long as `continue_training` is set to true in the main configuration).

`continue_training`, however, begins training from the last state. Passing checkpoints allows for more granularity from which state to start training (+ allowing the dynamic adjustment of hyperparameters which is not possible while continuing training).

But i am totally open to changing `skip_last_layer` to `True` by default, such that this has to be explicitly disabled, what do you prefer?

autrainer/training/utils.py
ramppdev: Sounds good, then i will change it to a try catch block for now, and later on find a more generalized solution (for e.g. fully convolutional networks) once we have a concrete use case.

docs/source/modules/models.rst
millinma: No strong opinion here. From a user perspective, the dangers of misuse are in the default `True` case that they believe they are continuing the training as they are not aware of the flag, which they are not and it would be obvious to know. In the default `False` case it might be that an error is being thrown because of the mismatch. I am aware of the `continue-training` option, which indeed kind of covers the same idea, but for instance is not usable if a run is cancelled prior to finishing say due to time limitations.

ramppdev: The main problem is when a user only passes an `optimizer_checkpoint` and there would be a potential shape mismatch, no error is thrown at all during loading the state dict. However in the first optimization step, there could be a potential shape mismatch for example when using Adam. This would then throw an error which seems completely unrelated with loading the optimizer checkpoint and is therefore completely masked.
Thats what my intention was to set `skip_last_layer` always to `True` in this case, preventing this issue from happening in the first place.

I do see the that there are dangers of misuse, but to discover checkpointing, a user has to read the docs anyways, so i feel like if we properly explain the `skip_last_layer` option, it should be fine either way.


For me, both `True` and `False` could be seen as intuitive, when considering that our "intended use case" of checkpointing is to first train a model on dataset A and later fine-tune it on dataset B, then `skip_last_layer` should be `True` by default, but still allowing the user to explicitly keep the same classification head if desired (and possible).
Right now, if the shape does not match, we also skip the layer, also if `skip_last_layer` is `False`. Should we keep this then if it is `True` by default or just fail?

What would you prefer?

ramppdev:
Ok i have greatly simplified the implementation here to mitigate hidden control flow that the user may not be aware of and added more clear documentation.

- allow the last layer to be both linear or convolutional. I did not add a try catch block since it will raise a `RuntimeError` anyways if loading fails here.
- `skip_last_layer` is now true by default, clearly distinguishing between continuing training and checkpointing with the intention of targeting a new task / dataset.
- documentation now clearly outlines the flaw with the current approach: we cannot accurately identify the last layer if it is not specified last in the modules (we had this issue before as well).

docs/source/modules/models.rst
ATriantafyllopoulos: I tend towards explicitly failing when something's not working rather than do something that the user might not want us to do

ramppdev: I do prefer this as well. Right now, we have `skip_last_layer: True` as default value, which is think is justified to distinguish checkpointing and continuing training. If the user explicitly sets `skip_last_layer: False` or no last layer can be identified, and this results in an actual shape mismatch, we just fail and not try to do some hidden control flow to recover.

millinma:
Okay, thanks a lot for all the consideration, seems like a good solution to me :)

ramppdev:
@millinma so should be ready to merge? I guess we could also release `v0.2.0` afterwards as we have quite the amount of new features + fixes.

millinma:
Yes, ready to merge I would say :)

ramppdev:
> Not a super expert on the versioning. So don't know if it should be 0.1.2 or 0.2.0

We should definitely follow [semantic versioning](https://semver.org/) to some extent, meaning major releases for incompatible changes, minor for feature releases, and patch for bug fixes only as soon as we reach `v1.0.0`.

Since we are below `v1.0.0`, i would stay with minor releases for all feature releases (including incompatible changes), and patch releases for fixes only.

This should be quite clear, as we are in a super early stage of development and things may change rapidly and drastically.
I would also collect some useful features and manually decide when a new release should be made, depending on how long the last release was made and how much new things are added, so we dont just spam releases.
However, for urgent fixes, i would instantly release a patch.

PR #21, opened 2024-10-04
Save Preprocessing Pipeline for Inference

### LINKED ISSUE #20: Save Preprocessing Pipeline for Inference
We currently only provide loading a custom preprocessing pipeline during inference with `-p`, however if the model is trained with e.g. spectrograms, the user has to provide this explicitly.

Therefore, the preprocessing pipeline should be saved alongside the training run so it can be loaded as the default preprocessing for inference.

ramppdev:
Closes #20 and additionally exports a `preprocess_file_handler.yaml` and `preprocess_pipeline.yaml` to be used during inference based on the `features_subdir` of the dataset.
If no preprocessing pipeline is used, `preprocess_file_handler.yaml` is the same as `file_handler.yaml`.

During inference `-p` / `preprocess-cfg` is now set to `default`, automatically loading the preprocessing pipeline used during training. If spectrograms were used for training, the user can still pass audio during inference and it is automatically transformed to spectrograms.
However, this can still be deactivated by setting the preprocessing config to `None` to disable preprocessing or to any local / global preprocessing pipeline. This essentially covers the old implementation.

I currently just "re-export" the preprocessing pipeline as an audobject during training, so that inference does not rely on Hydra (we dont have a Hydra instance during inference). This could also be done during preprocessing, and during training just copy `preprocess_file_handler.yaml` and `preprocess_pipeline.yaml` from the respective `features_subdir` if it exists, otherwise export the default YAML files.
However, i thought that the current version had less hidden control flow.

Some time in the future we should think about improving the performance of the trainer init, as it gets more and more tasks now. Saving more audobjects definitely adds a couple seconds, but is maybe parallelizable.

### DIFF
--- a/autrainer/core/scripts/inference_script.py
+++ b/autrainer/core/scripts/inference_script.py
@@ -130,12 +130,14 @@ class InferenceScript(AbstractScript):
             type=str,
             metavar="P",
             required=False,
-            default=None,
+            default="default",
             help=(
                 "Preprocessing configuration to apply to input. Can be a path "
                 "to a YAML file or the name of the preprocessing configuration "
                 "in the local or autrainer 'conf/preprocessing' directory. "
-                "If None, no preprocessing will be applied. Defaults to None."
+                "If 'default', the default preprocessing configuration used "
+                "during training will be applied. If 'None', no preprocessing "
+                "will be applied. Defaults to 'default'."
             ),
         )
         self.parser.add_argument(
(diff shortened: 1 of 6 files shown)


ATriantafyllopoulos (approved):
Thank you! Yes, the `__init__` has gotten quite bloated already, but I think it's fine in those early stages, we can refactor for code clarity later.

PR #33, opened 2024-10-14
Add Inference Progress Bars

ramppdev:
This adds numerous progress bar improvements:
-  added tqdm progress bars to both `Inference.predict_directory` and `Inference.embed_directory`, controllable via the `-f, --frequency` argument. If set to 0, the progress bar is disabled.
-  merged the `pbar_frequency` and `silent` attributes to `frequency` for `autrainer preprocess` with the same functionality as well such that the CLI and Python wrappers are consistent for all commands / functions.
- adjusted the `frequency` to 1 by default, which should align more with what a user would expect by default.

### DIFF
--- a/autrainer/core/scripts/inference_script.py
+++ b/autrainer/core/scripts/inference_script.py
@@ -20,6 +20,7 @@ class InferenceArgs:
     extension: str
     recursive: bool
     embeddings: bool
+    frequency: int
     preprocess_cfg: Optional[str]
     window_length: Optional[float]
     stride_length: Optional[float]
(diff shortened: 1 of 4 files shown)


autrainer/core/scripts/inference_script.py
```diff
@@ -123,7 +124,18 @@ def add_arguments(self) -> None:
                 "Defaults to False."
             ),
         )
-
+        self.parser.add_argument(
+            "-f",
+            "--frequency",
+            type=int,
```
ATriantafyllopoulos, line 130: You can add bool arguments via:

```python
a = argparse.ArgumentParser()
a.add_argument("-f", "--frequency", default=False, action="store_true")
...
foo = a.parse_args()
```
And then do "foo.f" to access it.

To avoid negative logic, if you really want to make progress bars active by default, I would name this to `"-s", "--silent"`.

ATriantafyllopoulos: Sorry, I was too hasty to comment. I see now that it also allows you to specify the number of steps..In that case, it makes sense to make it an `int` of course.

BUT I really don't think it's a good idea to name it `frequency` for an AUDIO toolkit ;)

ramppdev: Good point, but we need to find an abbreviation that works, -p is already taken for inference.

ramppdev: Changed to `-u / --update-frequency` for preprocessing, inference, and serving in https://github.com/autrainer/autrainer/pull/33/commits/060786dd6b9f75d1a87b969b4ce52a7cfa28c630.

PR #46, opened 2024-10-21
Add MSP-Podcast dataset

ATriantafyllopoulos:
Adds MSP-Podcast dataset. As I haven't done that in a while, it would be fantastic if you can have a look @ramppdev 

Additionally, I remove a stray print in `AIBO`. That was only there for debugging.

---
P.S. I am also wondering what's the preferred way for debugging a dataset when developing? For now, I used a `python` script:

```python
from autrainer.datasets.msppodcast import MSPPodcast


data = MSPPodcast(
    path="",
    index_column="FileName",
    target_column="EmoValf",
    file_type="wav",
    file_handler="autrainer.datasets.utils.AudioFileHandler",
    features_subdir="Audios",
    metrics=["autrainer.metrics.Accuracy"],
    tracking_metric="autrainer.metrics.Accuracy",
    batch_size=1,
    seed=42,
    categories=["A", "H", "N", "S"]
    # train_transform=autrainer.transforms.RandomCrop
)
train_loader = data.train_loader
print(data.target_transform)

print(next(iter(train_loader)))
```

But is there some command that we can use to quickly test the dataset? Or is it something we could plausibly add as another CLI option, e.g. `autrainer debug dataset`, which would then output some basic checks?

### DIFF
--- a/autrainer-configurations/dataset/MSPPodcast-EmoAct-wav.yaml
+++ b/autrainer-configurations/dataset/MSPPodcast-EmoAct-wav.yaml
@@ -1,0 +1,29 @@
+# Important: should be used with inference_batch_size: 1
+id: MSPPodcast-EmoAct-wav
+_target_: autrainer.datasets.MSPPodcast
+
+path: data/MSPPodcast
+features_subdir: Audios
+index_column: FileName
+target_column: EmoAct
+file_type: wav
+file_handler: autrainer.datasets.utils.AudioFileHandler
+
+criterion: autrainer.criterions.MSELoss
+metrics:
+  - autrainer.metrics.PCC
+  - autrainer.metrics.CCC
+  - autrainer.metrics.MSE
+  - autrainer.metrics.MAE
+tracking_metric: autrainer.metrics.CCC
+
+transform:
+  type: raw
+  base:
+    - autrainer.transforms.Expand:
+        size: 48000
+        axis: -1
+  train:
+    - autrainer.transforms.RandomCrop:
+        size: 48000
+        axis: -1
--- a/autrainer-configurations/dataset/MSPPodcast-EmoAct-wav.yaml
+++ b/autrainer-configurations/dataset/MSPPodcast-EmoAct-wav.yaml
@@ -3,7 +3,6 @@ id: MSPPodcast-EmoAct-wav
 _target_: autrainer.datasets.MSPPodcast
 
 path: data/MSPPodcast
-features_subdir: Audios
 index_column: FileName
 target_column: EmoAct
 file_type: wav
--- a/autrainer-configurations/dataset/MSPPodcast-EmoClass-big4-wav.yaml
+++ b/autrainer-configurations/dataset/MSPPodcast-EmoClass-big4-wav.yaml
@@ -3,7 +3,6 @@ id: MSPPodcast-EmoClass-big4-wav
 _target_: autrainer.datasets.MSPPodcast
 
 path: data/MSPPodcast
-features_subdir: Audios
 index_column: FileName
 target_column: EmoClass
 file_type: wav
--- a/autrainer-configurations/dataset/MSPPodcast-EmoClass-big4-wav.yaml
+++ b/autrainer-configurations/dataset/MSPPodcast-EmoClass-big4-wav.yaml
@@ -1,0 +1,30 @@
+# Important: should be used with inference_batch_size: 1
+id: MSPPodcast-EmoClass-big4-wav
+_target_: autrainer.datasets.MSPPodcast
+
+path: data/MSPPodcast
+features_subdir: Audios
+index_column: FileName
+target_column: EmoClass
+file_type: wav
+file_handler: autrainer.datasets.utils.AudioFileHandler
+
+criterion: autrainer.criterions.BalancedCrossEntropyLoss
+metrics: 
+  - autrainer.metrics.Accuracy
+  - autrainer.metrics.UAR
+  - autrainer.metrics.F1
+tracking_metric: autrainer.metrics.Accuracy
+
+categories: A,H,N,S
+
+transform:
+  type: raw
+  base:
+    - autrainer.transforms.Expand:
+        size: 48000
+        axis: -1
+  train:
+    - autrainer.transforms.RandomCrop:
+        size: 48000
+        axis: -1
--- a/autrainer-configurations/dataset/MSPPodcast-EmoClass-big4-wav.yaml
+++ b/autrainer-configurations/dataset/MSPPodcast-EmoClass-big4-wav.yaml
@@ -15,7 +15,7 @@ metrics:
   - autrainer.metrics.F1
 tracking_metric: autrainer.metrics.Accuracy
 
-categories: A,H,N,S
+categories: [A,H,N,S]
 
 transform:
   type: raw
--- a/autrainer/datasets/msppodcast.py
+++ b/autrainer/datasets/msppodcast.py
@@ -1,0 +1,202 @@
+import os
+from functools import cached_property
+from typing import Dict, List, Optional, Tuple, Union
+
+from omegaconf import DictConfig
+import pandas as pd
+import torch
+
+from autrainer.transforms import SmartCompose
+
+from .abstract_dataset import AbstractDataset
+from .aibo import Standardizer
+from .utils import (
+    AbstractTargetTransform,
+    LabelEncoder,
+    MinMaxScaler,
+)
+
+class MSPPodcast(AbstractDataset):
+    def __init__(
+        self,
+        path: str,
+        features_subdir: str,
+        seed: int,
+        metrics: List[str],
+        tracking_metric: str,
+        index_column: str,
+        target_column: str,
+        file_type:str,
+        file_handler: Union[str, DictConfig, Dict],
+        batch_size: int,
+        inference_batch_size: Optional[int] = None,
+        train_transform: Optional[SmartCompose] = None,
+        dev_transform: Optional[SmartCompose] = None,
+        test_transform: Optional[SmartCompose] = None,
+        stratify: Optional[List[str]] = None,
+        standardize: bool = False,
+        categories: List[str] = None,
+    ) -> None:
+        """MSP-Podcast dataset.
+
+        .. warning::
+            There are multiple versions available for this dataset.
+            We recommend always using the latest one 
+            (v1.11 at the time of writing)
+            but our code is set up to work with all versions
+            (at least up to v1.11).
+        
+        .. note::
+            Note that after v1.7, the dataset features two test sets.
+            We only use ``test1'', as ``test2'' was found to be biased
+            with respect to gender. 
+            See https://doi.org/10.21437/Interspeech.2019-1708.
+
+        .. note::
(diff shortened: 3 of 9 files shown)


autrainer-configurations/dataset/MSPPodcast-EmoAct-wav.yaml
```diff
@@ -0,0 +1,29 @@
+# Important: should be used with inference_batch_size: 1
+id: MSPPodcast-EmoAct-wav
+_target_: autrainer.datasets.MSPPodcast
+
+path: data/MSPPodcast
+features_subdir: Audios
```
ramppdev, line 6: This would need to be changed to `features_subdir: default` for all configurations that load the dataset in its default state (ref [Structure and Loading](https://autrainer.github.io/autrainer/modules/datasets.html)).

autrainer/datasets/aibo.py
```diff
@@ -96,7 +96,6 @@ def __init__(
         self.standardize = standardize
         if self.standardize:
             train_data = torch.cat([x for x, *_ in self.train_dataset])
-            print(train_data.mean(0).shape, train_data.std(0).shape)
```
ramppdev, line 99: Thanks! Missed that.

autrainer/datasets/msppodcast.py
```diff
@@ -0,0 +1,202 @@
...
+        As this dataset is not publicly-available, please download it manually
+        by contacting Prof. Carlos Busso: 
+        https://ecs.utdallas.edu/research/researchlabs/msp-lab/MSP-Podcast.html
+
+        This function will not do anything.
+
+        For more information on the data, see:
+        https://doi.org/10.1109/TAFFC.2017.2736999
+        """
+        return None
```
ramppdev, line 145: The download method could be used to (ref [Structure and Loading](https://autrainer.github.io/autrainer/modules/datasets.html)) transform the data from the original state it is provided to our "opinionated" format, this could be similar to AIBO which is also private:
https://github.com/autrainer/autrainer/blob/4dd050d87e27e71c676da9788e2e4c5a29aaed17/autrainer/datasets/aibo.py#L124-L196

This would be required as we make the following assumptions:
- `${path}/default`: the default directory holds the raw files (audio)
- `${path}/*.csv`: the index (file paths) and targets are mapped through one or more dataframes

autrainer/datasets/msppodcast.py
```diff
@@ -0,0 +1,202 @@
...
+    ) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
+        """Load the dataframes.
+
+        Overrides base class so there is no need
+        to create separate train/dev/test files.
+
+        Returns:
+            Dataframes for training, development, and testing.
+        """
+        df = pd.read_csv(os.path.join(self.path, "Labels", "labels_consensus.csv"))
```
ramppdev, line 158: All CSVs of the dataset have to be located directly in the `${path}` directory. But its totally fine to construct the subsets from `labels_consensus.csv`.

autrainer/datasets/msppodcast.py
```diff
@@ -0,0 +1,202 @@
...
+    def output_dim(self) -> int:
+        """Get the output dimension of the dataset.
+
+        Determined automatically based on the type of task.
+
+        Returns:
+            Number of classes.
+        """
+        if self.task == "classification":
+            return len(self.df_train[self.target_column].unique())
```
ramppdev, line 198: Idea (not for here, just general): It may be a good idea to implement `__len__` for all target transforms, this could be more intuitive, do you agree?

ramppdev (commented):
Overall the dataset is in a very good shape already, thanks!

But as we enforce a pretty opinionated dataset structure (this is needed such that we can preprocess the datasets), the paths of the raw audio files and of the metadata CSV need to be adjusted.

Also, we need to check that the actual `index_colum` file paths in the CSV are relative to the `${path}/${features_subdir}` directory.

autrainer/datasets/msppodcast.py
```diff
@@ -0,0 +1,202 @@
...
+
+        Returns:
+            Target transform.
+        """
+        if self.task == "classification":
+            return LabelEncoder(
+                self.df_train[self.target_column].unique().tolist()
+            )
+        elif self.task == "regression":
+            return MinMaxScaler(
```
ramppdev, line 181: If #47 is merged, the target column should be passed to the `MinMaxScaler` now.

autrainer-configurations/dataset/MSPPodcast-EmoAct-wav.yaml
ATriantafyllopoulos: Hmm, so how would you recommend doing it since the original download will create an "Audios" folder? I would make a symlink for me, but I'm not sure if that works for windows and would want to avoid a copying..

It's one of the things I like the least about our current state. This is only needed for preprocessing, right?

autrainer/datasets/msppodcast.py
ATriantafyllopoulos: Again, as above, this is for the `preprocessing` transform, right? And you would basically add instructions for the user and expect them to rename the `Audios` folder to `default`?

autrainer/datasets/msppodcast.py
ATriantafyllopoulos: You mean I need to save them for the preprocessing to work, right?

autrainer/datasets/msppodcast.py
ATriantafyllopoulos: Yes, absolutely. Please open a new issue to track and resolve this

ramppdev: Ref #49

autrainer-configurations/dataset/MSPPodcast-EmoAct-wav.yaml
ramppdev: Is it possible to move the data? If so i would say it makes sense to keep all datasets in the same structure and explain the transformation. Also ref https://github.com/autrainer/autrainer/pull/46#discussion_r1812740439

ATriantafyllopoulos: Usually (at least for our infrastructure), it should be avoided, because not all users have write access. Also, more people might work with `MSP-Podcast` (or other data) without using `autrainer`. In that case, we don't want to destroy the original structure  for them

ramppdev: In this case it is totally fine to symlink the dataset stucture by creating the `features_subdir` (as a symlink to the Audio dir) and also symlink the CSV (if the filepaths need adjustment its best to copy them).
However this would mean that preprocessed directories wont be located in the "globally shared" dataset location but in `${path}` next to the `${path}/default/` symlinked directory.

For all other users, the description of the download method (accessible via the docs) could outline the required manual preprocessing.
It is also possible that you provide a download method that transforms the dataset as needed (from the downloaded format to ours), and then just do the symlink yourself without using the download method. After the dataset root directory exists, the download method wont re-run and you will still be able to preprocess the dataset.

autrainer/datasets/msppodcast.py
ramppdev: The way to circumvent this is to add additional attributes to all dataset configurations specifying the paths of the default dataset subdirectory (relative to `${path}` or absolute) and providing a list of CSV paths in the same way.

Thats how completely customized paths could be used in theory, but i originally opted against this and "enforced" a predefined dataset structure as dataset configurations are already quite complicated and this would add even more things for users to consider and set up.

autrainer-configurations/dataset/MSPPodcast-EmoClass-big4-wav.yaml
```diff
@@ -0,0 +1,30 @@
...
+file_handler: autrainer.datasets.utils.AudioFileHandler
+
+criterion: autrainer.criterions.BalancedCrossEntropyLoss
+metrics: 
+  - autrainer.metrics.Accuracy
+  - autrainer.metrics.UAR
+  - autrainer.metrics.F1
+tracking_metric: autrainer.metrics.Accuracy
+
+categories: A,H,N,S
```
ramppdev, line 19: The categories would be passed as the string `"A,H,N,S"` here, so this needs to be changed to `categories: [A,H,"N",S]`.

ramppdev:
Oh and 2 general things:
- i think your ruff setup is different as the file was not correctly formatted, e.g. you could just use the pre-commit hook to check that: https://autrainer.github.io/autrainer/development/contributing.html#conventions
- the dataset should also be added to `test_datasets.py`

autrainer/datasets/msppodcast.py
ramppdev: `output_dim` can now be safely removed and is handled by the `AbstractDataset` automatically.

autrainer/datasets/msppodcast.py
```diff
@@ -0,0 +1,202 @@
...
+            test_transform: Transform to apply to the test set.
+                Defaults to None.
+            stratify: Columns to stratify the dataset on. Defaults to None.
+            standardize: Whether to standardize the data. Defaults to False.
+            categories: used to filter out specific emotional categories. 
+                Useful for training on subset of data/classes, such as the classic
+                ["A", "H", "N", "S"] 4-class problem found in literature. 
+                Defaults to None.
+        """
+        # TODO: expand for multitask regression
```
ramppdev, line 90: Multi-target regression is proposed in #53.

ATriantafyllopoulos:
Finally finished the refactoring :)

Will wait until #111 is fixed so I adopt the new format too, and then we should be good to go

autrainer/datasets/msppodcast.py
```diff
@@ -0,0 +1,212 @@
...
+            Dataframes for training, development, and testing.
+        """
+        df = pd.read_csv(
+            os.path.join(self.path, "Labels", "labels_consensus.csv")
+        )
+        if self.categories is not None:
+            df = df.loc[df["EmoClass"].isin(self.categories)]
+        return df.reset_index(drop=True)
+
+    @cached_property
```
ramppdev, line 147: i think the above could optionally be simplified to `def df(self) -> pd.DataFrame:`, and then:
```python
@cached_property
def df_train(self):
        return self._df.loc[df["Split_Set"] == "Train"]
```
and optionally without dropping the index, as we handle this here: https://github.com/autrainer/autrainer/blob/d558da2cb01aedb43546b0f42ab10f99dd7ba1d5/autrainer/datasets/utils/dataset_wrapper.py#L77

ramppdev (approved):
Looking great!

autrainer/datasets/msppodcast.py
ATriantafyllopoulos: Nah, I'll keep the `_load_df` as property 

I think it's important that `drop=True` to avoid confusion with the indices in the stored `csvs`

ramppdev:
HF model is now up to speed: https://huggingface.co/autrainer/msp-podcast-emo-class-big4-w2v2-l-emo/commit/f3308108b72eb04742774c301d0c2170bb85ad85

PR #47, opened 2024-10-23
Save Class Probability Columns in Outputs Tracker and Inference Results

### LINKED ISSUE #39: Save Class Probabilities during Inference
During inference prediction, the `results.csv` outputs the filename, optional offset (sliding window), prediction, and output (list of logits) for each sample.
This should be extended to include the output probabilities as well.
- classification: include a separate column for each class in the dataset
- regression: include a separate column for each target (re-scaled): this requires the target transform to have names for the target similar to classification. We currently do not yet support multi-target regression but this may be good to add then as well.

ramppdev:
Closes #39 by tracking additionally saving the class probabilities alongside the model outputs / predictions.
- Split target transforms into separate files for readability.
- Extended the target transform pipeline: `M.outputs` -> `T.probabilities_batch` -> `T.predict_batch` -> `T.decode` (elementwise) -> `decoded predictions`
  - Instead of model output logits, `T.predict_batch` now takes in model output probabilities
  - While being unnecessrary, also applying the Softmax function for multi-class classification now as it is technically correct and unifies the pipeline purpose
  - For regression, `T.probabilities_batch` is a no-op.
- To obtain a mapping of the target / labels and the probabilities, added `T.probabilities_to_dict` (this means that the `MinMaxScaler` now also adds the `target_column` as its target name).
- Based on `T.probabilities_to_dict`, a column for each target / label can be constructed. The columns are added both to the `dev_results.csv` produced during training for each evaluation, `test_results.csv` for testing, and `results.csv` produced during inference.
- Caught a small bug where audio files were padded during inference without knowing their sample rate.


autrainer/datasets/utils/target_transforms/min_max_scaler.py
```diff
@@ -0,0 +1,97 @@
...
+        """Get the encoded probabilities from a batch of model outputs by
+        returning the raw values.
+
+        Args:
+            x: Batch of model outputs.
+
+        Returns:
+            Encoded probabilities.
+        """
+        return x
```
ATriantafyllopoulos, line 61: Shouldn't this be put through `sigmoid`?

ramppdev: Yeah, for the regression task there is no way of knowing that the output lies in this range, i.e. we now treat everything as raw logits and as there is no adequate "probability" function for the continuous regression targets `probabilities_batch` serves as a no-op here.

Why should we apply sigmoid here?

ATriantafyllopoulos: Because the `MinMaxScaler` conceptually serves to limit the target range to `[0, 1]`. However, since the training happens with raw logits, the model can still output values beyond that range, which are then transformed back to the original scale.

An example is age classification. We might train a model to predict age values in the range `[0, 100]`, and use the `MinMaxScaler` to limit that to `[0, 1]` for training. But we don't want the model to output anything `<0`

ramppdev: Ok yeah so the `probabilities_batch` should take care of assuring that the expected limit is actually enforced? This raises 2 questions for me:
1. Why do we want to appy the sigmoid function here? Since we train the model on the "raw logits" wouldnt this conceptually change the actual "prediction" we present to the user in a non-linear way? Also i assume [0,1] may not be the only range used for scaling? Would just clamping the output to the range [minimum, maximum] solve the issue more elegantly?
2. What happens if the user does not want to scale at all? We should probably provide another target transform for regression in that case?

ATriantafyllopoulos: 1. Yes, it was actually why the `sigmoid` flag was there in some models, now that I think back on it ;)

2. Yeah, this is a question of how much we hide under the hood. The user now provides a column that needs to be regressed. We assume that this column covers the entire range of the values that the user expects to get in the output (e.g. age in the range 0-100). I think from a user perspective, I would expect to get the same range in the output (or less of course, if the model never predicts correctly) but *nothing* outside that range. And this can be achieved by either a `sigmoid` or `clamping`. I find the `sigmoid` a more natural, continuous and smooth transformation, but maybe `clamping` also works (though it might harm correlations)..

ramppdev: Okay makes sense now when the sigmoid is included in the training process as well.
If we train on the raw logits it would make more sense to me to just clamp the values IMO.
I still dont get why it would be fine to transform the outputs in a non-linear way using the sigmoid function only for inference without the model training on it?

ATriantafyllopoulos: Oh yes, I was thinking mainly in terms of Spearman's rho, which won't change after applying a monotonic transformation, but of course MSE and stuff would break.

So do you think we should clamp by default without any explicit user control? It's also not my favorite. 

I guess the optimal would be to always use `sigmoid` when doing regression (and then always combine with the `MinMaxScaler` and make the sigmoid part of the model). Do we want to reintroduce that? But that will then "allow" the user to use `sigmoid` also for `multi-label`, where we deactivated it to prevent doing it twice.

ramppdev: One solution avoiding re-introducing potentially problematic sigmoid / softmax layers in the forward pass would be:
- Since the `MinMaxScaler` has targets in the range of [0,1] it is safe to apply the Sigmoid function in `probabilities_batch` during **both** training and inference -> call `probabilities_batch` before passing the outputs to the loss function.
- Loss functions now take in probabilities: this would mean switching to from `CrossEntropyLoss` to `NLLLoss` for multi-class and from `BCEWithLogitsLoss` to `BCELoss` for multi-label classification as `probabilities_batch` applies the corresponding Softmax / Sigmoid layer already.
- At the same time allow different choices for target transforms in the case of (multi-target) regression: introduce a new parameter to regression dataset configurations, selecting either "bounded" (our `MinMaxScaler`) or "unbounded" (new simplified target transform without scaling and without applying sigmoid) regression.

The other solution would be to not include Sigmoid during inference as the network learns to model the targets based on its raw logit output. After that we can clamp the outputs (here it is also a question if the user wants that, maybe introducing a new simplified regression target transform could be useful here as well), which will also result in a bounded  prediction in the expected range, and correctly represents the model outputs.

Both solution are valid in my opinion, the first one provides more control to the user in my opinion but is not backwards compatible. I would avoid introducing sigmoid / softmax layers in models as this can potentially lead to more complex bugs downstream which are definitely harder to debug / constrain through checks everywhere (we didnt notice the potential of the bug for quite some time).

While we are at it and talking about regression target scaling: Should we calculate the minimum and maximum based on the complete dataset targets not only train?
https://github.com/autrainer/autrainer/blob/0d27903f7b3d889b4cd40fbe07d053424ac20db8/autrainer/datasets/abstract_dataset.py#L545-L551

ATriantafyllopoulos: > While we are at it and talking about regression target scaling: Should we calculate the minimum and maximum based on the complete dataset targets not only train?

No, we should not introduce more information leakage from other sets :)

> we could also just add a new argument to probabilities_batch like train: False which dynamically determines if the model is in training or eval mode, and have a no-op for classification tasks during training here.

I find that way too opaque for users to follow, and would avoid that. I'd rather prefer an option that can be easily made transparent to the users (and potential contributors)

The problem with the loss functions for classification is that `NLLloss` needs `log_softmax`, not "just" `softmax`.

---

Let me try to think out loud. We're discussing here an interface between several things:

1. The labels, and correspondingly the dataset, which determine the range for the target value
2. The model which learns to predict this target value, and should ideally only output values in the input range (note: for classification, let's assume our range is [0, 1] for each logit/probabilities)
3. The loss function, which might be "opinionated" regarding what goes in and out (e.g. `NLLLoss` needs `log_softmax` inputs)

So the `TargetTransform` should handle all of this. For now, we have chosen to use the `TargetTransform` only under the hood, i.e. to hide it from the user. I think we should keep it like this.

The user then expects:

a. For regression, to get one column with values in the same range as what they gave as targets
b. For classification, one column per class with its probability (plus one column with the argmax prediction in case of standard classification)

The `TargetTransform` is "fitted" inside the dataset, thus should not have any information regarding the model or loss function. This is why we disabled the option to use `sigmoid/softmax` in the models.

It feels to me as if the `TargetTransform` should have two separate methods for use during training (i.e. what is passed to the loss function) and inference (i.e. what is tracked during training and exposed to the user).

+ For `LabelEncoder`, the training method would be a no-op and the inference method a `softmax`
+ For `MinMaxScaler`, both methods would apply the `sigmoid`

How do you think about that?

ramppdev: This sounds like a very good solution to me!
This is basically the better approach of the previous solution but disentangles the if training / else check and rightfully delegates that to separate more clear functions. I will get to work.

Regarding the bounded / unbounded regression: should we also introduce an additional target transform which does not apply the Sigmoid function and target scaling?

ATriantafyllopoulos: > Regarding the bounded / unbounded regression: should we also introduce an additional target transform which does not apply the Sigmoid function and target scaling?

I would not do it for now because I don't want to expose it as an option to the user yet. It's not yet clear how to do that properly. E.g. now the choice of transform is based on the dataset class, but then it would become conditional on some hparam from `hydra`, and the user might not know where to start. Let's not give too many freedoms yet ;)

ramppdev:
Alright, implemented https://github.com/autrainer/autrainer/pull/47#discussion_r1816725188 as follows:
- renamed `probabilities_batch` and `predict_batch` to `probabilities_inference` and `predict_inference`
- added the Sigmoid function to `MinMaxScaler.probabilities_inference` such that outputs are bound to the range [0,1] during inference
- added `probabilities_training` to all target transforms (no-op for classification, Sigmoid for regression)
- added `probabilities_training` between the model outputs and criterion for training and evaluation (during evaluation needed to obtain the correct loss)
- updated optional `optimizer.custom_step` to use the `probabilities_training` function

In any case we still report the model output "logits" and only use the "probabilities" for the loss function i.e. everything stays the same in terms of logged outputs.

ATriantafyllopoulos (approved):

PR #53, opened 2024-10-27
Add Multi-Target Regression Task

### LINKED ISSUE #48: Add Multi-Target Regression Task
We currently support multi-class and multi-label classification but only single-target regression.
It is useful to add this, as #46 would also benefit from using multiple regression targets at once.

ramppdev:
Closes #48 by adding multi-target regression:
- added new `MultiTargetMinMaxScaler` which applies the `MinMaxScaler` for each target column separately
- added new base dataset
- extended `ToyDataset` to multi-target regression (including configurations)
- extended regression metrics to handle multiple targets by calculating the metric for each target and averaging


ATriantafyllopoulos (approved):
Phenomenal! Thanks a million for taking care of this!

PR #54, opened 2024-10-29
Update dataset interface

ATriantafyllopoulos:
@ramppdev this is a draft PR for changing the underlying dataset. I want your feedback on whether I'm on the right direction before finishing/debugging/testing.

Basically I have done two major changes:

1. Use `@cached_property` for `train_df`, `dev_df`, `test_df`, which defaults to the current structure, but can be overriden if the user doesn't want to change the original format. I also adapted the datasets already which *cannot* be downloaded automatically. For those that can, I have no problem with the original format.ee 

2. Change the `preprocess_script` to *instantiate* the dataset rather than load the CSVs manually, then rely on the `DatasetWrapper` and `torch.utils.data.DataLoader` interfaces for iterating through the data, extracting features, and storing the data

There's probably a few other things to change, but hopefully this would relax the restrictions we have on the "hard" dataset format and rely on the programmatic interface of `DatasetWrapper`..

So let me know what you think of this new format and whether I've messed up something greatly before I continue!

### DIFF
--- a/autrainer/core/scripts/preprocess_script.py
+++ b/autrainer/core/scripts/preprocess_script.py
@@ -145,8 +145,6 @@ class PreprocessScript(AbstractPreprocessScript):
             features_subdir = dataset["features_subdir"]
             dataset["features_subdir"] = None
             data = autrainer.instantiate(dataset)
-            if features_path is None:
-                features_path = data.path
             # manually disable dataset transforms
             data.train_transform = None
             data.dev_transform = None
--- a/autrainer/core/scripts/preprocess_script.py
+++ b/autrainer/core/scripts/preprocess_script.py
@@ -160,7 +160,7 @@ class PreprocessScript(AbstractPreprocessScript):
                     dataset=d,
                     shuffle=False,
                     num_workers=self.num_workers,
-                    batch_size=1,  # TODO: can we do it batched?
+                    batch_size=1,
                 )
                 for data in tqdm.tqdm(
                     loader,
--- a/autrainer/core/scripts/preprocess_script.py
+++ b/autrainer/core/scripts/preprocess_script.py
@@ -144,7 +144,6 @@ class PreprocessScript(AbstractPreprocessScript):
             data.train_transform = None
             data.dev_transform = None
             data.test_transform = None
-            os.makedirs(data.features_subdir, exist_ok=True)
 
             pipeline = SmartCompose(
                 [
--- a/autrainer/core/scripts/preprocess_script.py
+++ b/autrainer/core/scripts/preprocess_script.py
@@ -139,8 +139,8 @@ class PreprocessScript(AbstractPreprocessScript):
             dataset["file_type"] = None
             dataset["seed"] = 0  # ignored
             dataset["batch_size"] = 8  # ignored
-            _ = dataset.pop("criterion")
-            _ = dataset.pop("transform")
+            dataset.pop("criterion")
+            dataset.pop("transform")
             features_path = dataset.pop("features_path")
             features_subdir = dataset["features_subdir"]
             dataset["features_subdir"] = None
--- a/autrainer/core/scripts/preprocess_script.py
+++ b/autrainer/core/scripts/preprocess_script.py
@@ -141,7 +141,7 @@ class PreprocessScript(AbstractPreprocessScript):
             dataset["batch_size"] = 8  # ignored
             dataset.pop("criterion")
             dataset.pop("transform")
-            features_path = dataset.pop("features_path")
+            features_path = dataset.pop("features_path", dataset["path"])
             features_subdir = dataset["features_subdir"]
             dataset["features_subdir"] = None
             data = autrainer.instantiate(dataset)
--- a/autrainer/core/scripts/preprocess_script.py
+++ b/autrainer/core/scripts/preprocess_script.py
@@ -108,7 +108,6 @@ class PreprocessScript(AbstractPreprocessScript):
             )
 
     def _preprocess_datasets(self) -> None:
-        from concurrent.futures import ThreadPoolExecutor
         import os
         from pathlib import Path
 
@@ -117,80 +116,6 @@ class PreprocessScript(AbstractPreprocessScript):
 
         from autrainer.datasets.utils import AbstractFileHandler
         from autrainer.transforms import SmartCompose
-
-        def _get_csvs(path: str) -> List[str]:
-            return [
-                f
-                for f in os.listdir(path)
-                if os.path.isfile(os.path.join(path, f)) and f.endswith(".csv")
-            ]
-
-        def _read_csvs(path: str, files: List[str]) -> List[pd.DataFrame]:
-            return [pd.read_csv(os.path.join(path, f)) for f in files]
-
-        def _get_unique_files(
-            dfs: List[pd.DataFrame],
-            column: str,
-        ) -> List[str]:
-            return pd.concat(dfs, ignore_index=True)[column].unique().tolist()
-
-        def _create_subdirs(path: str, files: List[str]) -> None:
-            dirs = {os.path.dirname(os.path.join(path, f)) for f in files}
-
-            for d in dirs:
-                os.makedirs(d, exist_ok=True)
-
-        def _split_chunks(files: List[str], chunks: int) -> List[List[str]]:
-            if chunks == 1:
-                return [files]
-            avg = len(files) / chunks
-            out = []
-            last = 0
-            while last < len(files):
-                out.append(files[int(last) : int(last + avg)])
-                last += avg
-            return out
-
-        def _process_chunk(
-            chunk: List[str],
-            dataset: dict,
-            preprocess: dict,
-            pbar: tqdm,
-            lock: Any,
-        ) -> None:
-            file_handler = autrainer.instantiate_shorthand(
-                config=preprocess["file_handler"],
-                instance_of=AbstractFileHandler,
-            )
-            output_file_handler = autrainer.instantiate_shorthand(
-                config=dataset["file_handler"],
-                instance_of=AbstractFileHandler,
-            )
-            pipeline = SmartCompose(
-                [
-                    autrainer.instantiate_shorthand(t)
-                    for t in preprocess["pipeline"]
-                ]
-            )
-
-            file_count = 0
-            for file_path in chunk:
-                in_path = Path(dataset["path"], "default", file_path)
-                out_path = Path(
-                    dataset["path"],
-                    dataset["features_subdir"],
-                    file_path,
-                ).with_suffix("." + dataset["file_type"])
-                data = file_handler.load(in_path)
-                data = pipeline(data, 0)
-                output_file_handler.save(out_path, data)
-                del data
-                file_count += 1
-                if file_count % self.update_frequency == 0:
-                    with lock:
-                        pbar.update(self.update_frequency)
-            with lock:
-                pbar.update(file_count % self.update_frequency)
 
         print("Preprocessing datasets...")
         for (name, dataset), preprocess in zip(
@@ -257,36 +182,8 @@ class PreprocessScript(AbstractPreprocessScript):
                         continue
                     output_file_handler.save(
                         out_path,
-                        pipeline(data[0], 0)  #TODO: why 0?
+                        pipeline(data[0], 0)
                     )
-            # csvs = _get_csvs(dataset["path"])
-            # dfs = _read_csvs(dataset["path"], csvs)
-            # unique_files = _get_unique_files(dfs, dataset["index_column"])
-            # _create_subdirs(
-            #     os.path.join(dataset["path"], dataset["features_subdir"]),
-            #     unique_files,
-            # )
-            # chunks = _split_chunks(unique_files, self.num_workers)
-            # lock = tqdm.get_lock()
-            # with tqdm(
-            #     total=len(unique_files),
-            #     desc=name,
-            #     disable=self.update_frequency == 0,
-            # ) as pbar:
-            #     with ThreadPoolExecutor(self.num_workers) as executor:
-            #         futures = [
-            #             executor.submit(
-            #                 _process_chunk,
-            #                 chunk,
-            #                 dataset,
-            #                 preprocess,
-            #                 pbar,
-            #                 lock,
-            #             )
-            #             for chunk in chunks
-            #         ]
-            #         for future in futures:
-            #             future.result()
 
 
 @catch_cli_errors
--- a/autrainer/core/scripts/preprocess_script.py
+++ b/autrainer/core/scripts/preprocess_script.py
@@ -25,6 +25,90 @@ class PreprocessArgs(PreprocessArgs):
     update_frequency: int
 
 
+def preprocess_main(
+    name: str,
+    dataset: DictConfig,
+    preprocess: DictConfig,
+    num_workers: int,
+    update_frequency: int,
+):
+    import os
+    from pathlib import Path
+
+    from tqdm import tqdm
+
+    from autrainer.datasets.utils import AbstractFileHandler
+    from autrainer.transforms import SmartCompose
+
+    print(f" - {name}")
+    if preprocess is None:
+        print("No preprocessing specified. Skipping...")
+        return
+    # set dataset file handler as output handler
+    # as dataset is already configured
+    # to work with the output of the preprocessing script
+    output_file_handler = autrainer.instantiate_shorthand(
+        config=dataset["file_handler"],
+        instance_of=AbstractFileHandler,
+    )
+    output_file_type = dataset["file_type"]
+
+    # override dataset file handling to work with raw audio
+    dataset["file_handler"] = preprocess["file_handler"]
+    # None allows dataset to work with all audio files
+    dataset["file_type"] = None
+    dataset["seed"] = 0  # ignored
+    dataset["batch_size"] = 8  # ignored
+    dataset.pop("criterion")
+    dataset.pop("transform")
+    features_path = dataset.pop("features_path")
+    if features_path is None:
+        features_path = dataset["path"]
+    features_subdir = dataset["features_subdir"]
+
+    dataset["features_subdir"] = None
+    data = autrainer.instantiate(dataset)
+    # manually disable dataset transforms
+    data.train_transform = None
+    data.dev_transform = None
+    data.test_transform = None
+
+    pipeline = SmartCompose(
+        [autrainer.instantiate_shorthand(t) for t in preprocess["pipeline"]]
+    )
+    for d, n in (
+        (data.train_dataset, "train"),
+        (data.dev_dataset, "dev"),
+        (data.test_dataset, "test"),
+    ):
+        loader = torch.utils.data.DataLoader(
+            dataset=d,
+            shuffle=False,
+            num_workers=num_workers,
+            batch_size=1,
+        )
+        for instance in tqdm(
+            loader,
+            total=len(loader),
+            desc=f"{name}-{n}",
+            disable=update_frequency == 0,
+        ):
+            # TODO: will be streamlined once we switch to dataclass
+            index = d.df.index[int(instance[2])]
+            item_path = d.df.loc[index, d.index_column]
+            out_path = Path(
+                features_path,
+                features_subdir,
+                os.path.basename(item_path),
+            ).with_suffix("." + output_file_type)
+            os.makedirs(os.path.dirname(out_path), exist_ok=True)
+            if os.path.exists(out_path):
+                continue
+            output_file_handler.save(
+                out_path, pipeline(instance[0].squeeze(dim=0), 0)
+            )
+
+
 class PreprocessScript(AbstractPreprocessScript):
     def __init__(self) -> None:
         super().__init__(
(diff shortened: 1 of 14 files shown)


ramppdev:
Ok i have some general questions on the intended functionality to understand it fully:
- the goal is to relax the dataset configuration such that the dataframes can be located anywhere + have any format (e.g. txt files for AIBO)? By default the CSVs are expected to be located in `{path}/train.csv` and so on?
- delegate loading and returning the dataframes to the dataset instance during preprocessing such that we do not rely on a fixed format

autrainer/core/scripts/preprocess_script.py
```diff
@@ -196,39 +197,96 @@ def _process_chunk(
...
+            data = autrainer.instantiate_shorthand(dataset)
+            loader = torch.utils.data.DataLoader(
+                torch.utils.data.ConcatDataset(
+                    data.train_dataset,
+                    data.dev_dataset,
+                    data.test_dataset
+                ),
+                shuffle=False,
+                num_workers=self.num_workers,
+                batch_size=1  #TODO: can we do it batched?
```
ramppdev, line 223: This one can be removed i think, seems like the loader below does the same.

autrainer/core/scripts/preprocess_script.py
```diff
@@ -196,39 +197,96 @@ def _process_chunk(
...
-                        for chunk in chunks
-                    ]
-                    for future in futures:
-                        future.result()
+            for d, n in (
+                (data.train_dataset, "train"),
+                (data.dev_dataset, "dev"),
+                (data.test_dataset, "test"),
+            ):
+                loader = torch.utils.data.DataLoader(
```
ramppdev, line 236: Yes, we can do it batched in theory. We can get the current batch size directly from the main configuration file as well.

autrainer/core/scripts/preprocess_script.py
```diff
@@ -196,39 +197,96 @@ def _process_chunk(
...
+                    out_path = Path(
+                        dataset["path"],
+                        features_subdir,
+                        os.path.basename(item_path),
+                    ).with_suffix("." + output_file_type)
+                    if os.path.exists(out_path):
+                        continue
+                    output_file_handler.save(
+                        out_path,
+                        pipeline(data[0], 0)  #TODO: why 0?
```
ramppdev, line 260: The `0` is the index of the sample in the dataset, which we previously could not have known but needs to be passed here as some transforms may rely on it. Since preprocessing should be "static" it is hardcoded here for determinism, ref: https://github.com/autrainer/autrainer/blob/e1ecb6c2a4e98ed6e00abd9d6423117bfdb9abbb/autrainer/transforms/smart_compose.py#L93-L108

But yeah this would also be greatly streamlined once we switch to a dataclass for transforms and datasets.

ramppdev (commented):
There are some things in the preprocessing script i am not sure would work right now so we need to probably adapt this a bit again.
But from a dataset perspective i really like the direction!

autrainer/core/scripts/preprocess_script.py
```diff
@@ -196,39 +197,96 @@ def _process_chunk(
...
+            output_file_handler = autrainer.instantiate_shorthand(
+                config=dataset["file_handler"],
+                instance_of=AbstractFileHandler,
+            )
+            output_file_type = dataset["file_type"]
+            # set features_subdir to None (defaults to audio)
+            # so that iteration takes place over raw data
+            dataset["features_subdir"] = None
+            dataset["file_handler"] = preprocess["file_handler"]
+            dataset["file_type"] = preprocess["file_type"]
```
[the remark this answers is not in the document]
ATriantafyllopoulos: Oh, okay here I jumped to conclusions without checking too deeply.

I assumed that in this place `dataset["file_handler"]` will already be set to e.g. `npy` for numpy arrays? So I need to set it to whatever the `preprocessing` handles. Since we only allow preprocessing of audio files, I can then set this to  `AudioFileHandler`?

Maybe a different question, I was not sure how the script "understands" what data/feature combinations to extract. Are the dataset `yamls` the ones included in the main `config.yaml`? So then they are already using the `npy` chain of transforms for features? Or are there different config files used for preprocessing?

What is the workflow?

autrainer/core/scripts/preprocess_script.py
```diff
@@ -196,39 +197,96 @@ def _process_chunk(
...
+            if preprocess is None:
```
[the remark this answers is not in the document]
ATriantafyllopoulos: But the check was there no? https://github.com/autrainer/autrainer/blob/2cdd8bb64ca6550a69a58e22bf291664c6d2a9f0/autrainer/core/scripts/preprocess_script.py#L199

autrainer/core/scripts/preprocess_script.py
```diff
@@ -196,39 +197,96 @@ def _process_chunk(
...
                 continue
-
-            csvs = _get_csvs(dataset["path"])
-            dfs = _read_csvs(dataset["path"], csvs)
-            unique_files = _get_unique_files(dfs, dataset["index_column"])
-            _create_subdirs(
-                os.path.join(dataset["path"], dataset["features_subdir"]),
-                unique_files,
+            data = autrainer.instantiate_shorthand(dataset)
```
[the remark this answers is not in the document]
ATriantafyllopoulos: > At a later point if the user wants to use preprocessed data with the full dataset (another config), preprocessing will not run again and there will be missing data.

This is also something this PR changes, I forgot to write it, but it's something I wanted to change anyway. Now the data preprocessing runs again, but omits *single instances* that have been extracted in the past.

The original intent of this change is to account for cases where the `preprocessing` crashes or is stopped prematurely, and the user does not to re-extract all features. But I think this will cover your concern as well?

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: Do all transforms work in batched mode though?

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: I thought so..So maybe I just leave it to 0 for now?

autrainer/core/scripts/preprocess_script.py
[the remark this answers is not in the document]
ramppdev: > I assumed that in this place `dataset["file_handler"]` will already be set to e.g. `npy` for numpy arrays? So I need to set it to whatever the `preprocessing` handles. Since we only allow preprocessing of audio files, I can then set this to `AudioFileHandler`?

We can exploit multiple things here, e.g. if we preprocess log-Mel-spectrograms, the dataset config has the `NumpyFileHandler` and the `file_type` is set to `npy` which is our target file type that we use in combination with the dataset file handler to correctly export the features.
Since we assume that we are dealing with audio files when trying to extract log-Mel-spectrograms, the `file_type` is not needed as `AudioFileHandler` can load multiple audio file formats and the exact format used is hardcoded in the default `index_column` of our dataframes (this would be the file format the dataset comes with).

> Maybe a different question, I was not sure how the script "understands" what data/feature combinations to extract. Are the dataset `yamls` the ones included in the main `config.yaml`? So then they are already using the `npy` chain of transforms for features? Or are there different config files used for preprocessing?

We have an input file handler tasked with loading the features and an output file handler tasked with saving the features during preprocessing.

The input file handler is specified in the preprocessing config (e.g., in `conf/log_mel_16k` for log-Mel-spectrograms feature extraction).
Here we assume that our input data are audio files stored in a file type compatible with our `AudioFileHandler`. The concrete file type is specified in the `index_colum` file endings of our dataframes (these were loaded directly before such that the dataset instance would not overlay the file type).

The output file handler is defined in the dataset configuration file (e.g., the `NumpyFileHandler` with `npy` as file type for already processed log-Mel spectrograms).
This file handler from the dataset configuration serves as the handler to save the processed features (transformed through the `pipeline` in the preprocessing configuration) with the `file_type` format specified in the dataset configuration.
At the same time this file handler is tasked with loading the features during training.

So the full workflow is the combination of the input file handler from the preprocessing config -> the feature extraction pipeline from the preprocessing config -> the output file handler from the dataset config -> stored features at location `{features_subdir}` (matches the name of the preprocessing config) -> the file handler specified in the dataset configuration -> preprocessed data ready for training

autrainer/core/scripts/preprocess_script.py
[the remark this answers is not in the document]
ramppdev: Sorry i was too unspecific, i mean that it cant be None here as you access e.g., `preprocess["file_handler"]` 2 lines above risking a `KeyError`.

autrainer/core/scripts/preprocess_script.py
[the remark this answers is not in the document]
ramppdev: Ok got it, yeah this would solve the issue. But then, preprocessing has to be run explicitly for ALL datasets used (we need to keep this in mind). Right now if datasets subsets would be used we process them once for all and any dataset configuration relying on already extracted features would not need to be re-run. I think this was not clear to the user anyways and repeated execution of `autrainer preprocess` would just have been a no-op anyways.

autrainer/core/scripts/preprocess_script.py
ramppdev: No our transforms only support a single feature, e.g. we would need to set our preprocessing pipeline during dataset instantiation for `train_transform`, `dev_transform`, and `test_transform`. Then the loader would already receive processed features ready to be stored (however this needs to account for the inference batch size then).

autrainer/core/scripts/preprocess_script.py
ramppdev: Yeah i would say so, the dataclass refactor will probably change this part anyways.

ramppdev:
> > * delegate loading and returning the dataframes to the dataset instance during preprocessing such that we do not rely on a fixed format
> 
> Yes, that's the main place where we need our fixed format, right? Ultimately, we also need this for datasets where we don't have (file, label) tuples but rather (segment, label) ones, i.e. a file contains multiple segments each with a different label.

Essentially, yes. We currently force the following format:
- all processed `{feature_subdir}`(s) to be present in `{path}` alongside all dataframes
- dataframes being CSVs with commas as separators etc.
- the `index_column` being relative to `feature_subdir`
- the default audio data being located in `{path}/default` and the audio having the exact file type specified in the actual CSVs as file endings (we overlay this using `file_type` when using preprocessed features)

We currently also dont support segments of a file, but we could add a transform to trim this in the future (with the dataclass refactor that could provide this additional information or sth similar).
But this is generally tricky when preprocessing features and the shape changes (e.g., time-based segmentation wont work then).
Our current solution was to just handle this in the dataset `download` method and crop / extract the segments, but we dont have such a dataset at the moment as an example.

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: Yes, I would avoid using the `train_transform` in particular, since it will most likely include a `RandomCrop`

ramppdev: This referred to using the actual preprocessing transform as the `train_transform` etc. (ignoring the ones in the dataset configuration). This way we dont have to manually apply the transform in the preprocessing script, can use the batch size in the main configuration, and potentially not require modification later on when we switch to a dataclass for feature management here.

ATriantafyllopoulos: Got it, yes that makes sense! I'm doing an initial refactoring now, let's see how that goes and maybe we adopt this afterwards

ATriantafyllopoulos: Okay, you can have a look at def0dda which should solve several of these issues by adding an `audio_subdir` parameter, which is used as `features_subdir` if that is set to `None`, and also solves the dataset instantiation issues you mentioned.

autrainer/core/scripts/preprocess_script.py
```diff
@@ -196,39 +197,96 @@ def _process_chunk(
...
+            # swap dataset handler with preprocessing handler
+            features_subdir = dataset["features_subdir"]
+            output_file_handler = autrainer.instantiate_shorthand(
+                config=dataset["file_handler"],
+                instance_of=AbstractFileHandler,
+            )
+            output_file_type = dataset["file_type"]
+            # set features_subdir to None (defaults to audio)
+            # so that iteration takes place over raw data
+            dataset["features_subdir"] = None
```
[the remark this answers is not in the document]
ATriantafyllopoulos: I tried to solve this by introducing an `audio_subdir` `parameter`, which in the `AbstractClass` is set to `default` and may be overridden in each derived class, e.g. I set it to `wav` for `AIBO`. Following that, I allow `features_subdir` to be `None`, and, if so, set it to this `audio_subdir`. The check now comes afterwards. This should allow us both to have a non-default structure, and to do the override to make the preprocessing script work.

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -17,9 +17,9 @@ def __init__(
         features_subdir: str,
         index_column: str,
         target_column: Union[str, List[str]],
-        file_type: str,
         file_handler: AbstractFileHandler,
         df: pd.DataFrame,
+        file_type: str = None,
```
ramppdev, line 22: It makes sense to not overlay the `file_type` if it should be the default one in the dataframe anyways. But for this to have the necessary effect we need in the preprocessing script (e.g., drop it from the dataset configuration to suppress overlying the type), we should make this explicitly nullable (`Optional[str]`).
This would mean that the `file_type` will no longer be required in dataset configurations without preprocessing.
Right now its kind of a hidden control flow that its nullable which could lead to interesting masked errors when relying on the file type when implementing custom datasets.

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -30,9 +30,12 @@ def __init__(
             features_subdir: Subdirectory containing the features.
             index_column: Index column of the dataframe.
             target_column: Target column of the dataframe.
-            file_type: File type of the features.
             file_handler: File handler to load the data.
             df: Dataframe containing the index and target column(s).
+            file_type: File type of the features. If `None`,
+                will not enforce a file_type. This can be useful
+                in case the dataset contains audio files
+                with different formats. Defualts to `None`.
```
ramppdev, line 38: typo `Defualts`

autrainer/core/scripts/preprocess_script.py
```diff
@@ -118,110 +117,36 @@ def _preprocess_datasets(self) -> None:
...
-                    data.train_dataset,
-                    data.dev_dataset,
-                    data.test_dataset
-                ),
-                shuffle=False,
-                num_workers=self.num_workers,
-                batch_size=1  #TODO: can we do it batched?
-            )
+            # manually disable dataset transforms
+            data.train_transform = None
```
ramppdev, line 145: I think at this point you could handle the pipeline in a batched fashion (https://github.com/autrainer/autrainer/pull/54#discussion_r1824249027) by just setting `data.train_transform = pipeline ` (pipeline below), and then just straight up save `data[0]` from the loader.

autrainer/core/scripts/preprocess_script.py
```diff
@@ -118,110 +117,36 @@ def _preprocess_datasets(self) -> None:
...
+            data.dev_transform = None
+            data.test_transform = None
+            os.makedirs(data.features_subdir, exist_ok=True)
```
ramppdev, line 148: We still want to enforce the location of the features subdir to be alongside the original path right (`{path}/log_mel_16k/spectrogram.npy`)?
This would be the path constructed right now (which at the same time allows us to conveniently check the matching preprocessing transform to apply).
This would lead to issues if the user has no write access to `{path}`.
The solution would be to override `audio_subdir` with an absolute path (pointing to the original dataset format, possible only with read access) as well as the dataframe properties to depend on `audio_subdir` as well.
Finally using `{path}` as the local place for feature extraction only.

Was this the intention that you had in mind?

autrainer/core/scripts/preprocess_script.py
ramppdev: i will close this as there is a comment for the new changes.

autrainer/datasets/aibo.py
```diff
@@ -110,16 +111,50 @@ def __init__(
...
-            ),
-            pd.read_csv(os.path.join(self.path, f"dev_{self.aibo_task}.csv")),
-            pd.read_csv(os.path.join(self.path, f"test_{self.aibo_task}.csv")),
+    @property
+    def audio_subdir(self):
+        """Subfolder containing audio data."""
+        return "wav"
+
+    @cached_property
+    def _load_df(self, path):
```
ramppdev, line 120: Oh and also here and below type annotations are missing and the `_load_df` call is missing `path`.

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: Do you foresee any unexpected errors when mapping the batched features to their original filenames using the returned index?

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: Ah, yes, another unforeseen complication. I'll think about it a bit, but I guess we might need to disentangle the two paths..

autrainer/core/scripts/preprocess_script.py
ramppdev: No this should be fine and also could be avoided by not shuffling the loader at all.

ATriantafyllopoulos: At second thought, this won't work unless we add `RandomCrop` since most of the time we have variable length sequences..

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: 75c07d8 added an optional `features_path` argument to `AbstractDataset`. Whenever it is provided, it overrides `self.path` as far as **features** are concerned (labels remain untouched).

There were some "hacks" to get this to work with `preprocess`, but I've finally been able to "test" this with EmoDB and it seems to work, both with the default `features_path=None` which stores features along with the original data, and with `features_path="features"` which uses a different path. The training also works with both.

This is a pretty big change under the hood though, which I'd like to have included in the tests. Do you have any idea how? Doesn't seem to be compatible with our `ToyDatasets` right now..

autrainer/core/scripts/preprocess_script.py
ramppdev: Yeah we missed that since the features can have any size here!

autrainer/core/scripts/preprocess_script.py
```diff
@@ -133,19 +132,25 @@ def _preprocess_datasets(self) -> None:
...
+
             # override dataset file handling to work with raw audio
             dataset["file_handler"] = preprocess["file_handler"]
             # None allows dataset to work with all audio files
             dataset["file_type"] = None
             dataset["seed"] = 0  # ignored
             dataset["batch_size"] = 8  # ignored
-            data = autrainer.instantiate_shorthand(dataset)
+            _ = dataset.pop("criterion")
+            _ = dataset.pop("transform")
```
ramppdev, line 143: I think these can just be popped without assigning to `_`.

autrainer/datasets/dcase_2018_t3.py
```diff
@@ -56,6 +57,10 @@ def __init__(
             file_type: File type of the features.
             file_handler: File handler to load the data.
             batch_size: Batch size.
+            features_path: Root path to features. Useful
+                when features need to be extracted and store
```
ramppdev, line 61: ...  when features need to be extracted and store`d` ... (for all new docstrings)

autrainer/core/scripts/preprocess_script.py
```diff
@@ -133,19 +132,25 @@ def _preprocess_datasets(self) -> None:
...
+            features_path = dataset.pop("features_path")
```
ramppdev, line 144: I think this would be more clear if we remove the if statement below and do `features_path = dataset.pop("features_path", dataset["path"])` instead? If you find it more complicated just keep it as is.

autrainer/core/scripts/preprocess_script.py
```diff
@@ -162,28 +167,27 @@ def _preprocess_datasets(self) -> None:
                     dataset=d,
                     shuffle=False,
                     num_workers=self.num_workers,
```
ramppdev, line 169: `num_workers` spawns subprocesses here for fetching, however we save examples in the main process again.
Before, we had separate threads handling the loading and saving operations, as i expect the feature extraction process to be mostly IO bound. I am not sure how we could re-use the same workers for saving the features separately, i.e. in the current setup i expect the preprocessing pipeline to run a bit slower.

autrainer/core/scripts/preprocess_script.py
ramppdev: Yeah it will be tricky to test with the `ToyDataset` and i would think these tests should be done for the base datasets where i have a small setup for mocking data already which you could reuse:
https://github.com/autrainer/autrainer/blob/2485ef339bcd90e96f4b63a0dedcb6c99358dd0c/tests/test_datasets.py#L80-L98

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: Yes, I know, I think it's fine if we keep the simpler version for now and optimize later, even though it is slower. I think the penalty might not be too great in the grand scheme of things, given that preprocessing happens once. We can also optimize it in a future PR

ramppdev: Sounds good, probably not too important as you say :)

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: This will fail if "features_path" is specified but is `None`. So I'll stick to my version

autrainer/core/scripts/preprocess_script.py
ATriantafyllopoulos: Added some tests in a0254a2

I also moved the core processing of the script to a new function, which made it easier to test.

This will also be useful if we provide a "proper" API in the future, as it's an entry-point for users to do the preprocessing in `python`

autrainer/datasets/dcase_2018_t3.py
ATriantafyllopoulos: de5c9b1

autrainer/datasets/utils/dataset_wrapper.py
ATriantafyllopoulos: Fixed

autrainer/datasets/aibo.py
```diff
@@ -110,16 +117,50 @@ def __init__(
                 self.df_train, self.train_transform
```
ATriantafyllopoulos, line 117: @ramppdev I am wondering if this is needed..

Since `self.train_dataset`, `self.dev_dataset`, and `self.test_dataset` are now `@cached_property`, they will be instantiated only once, when the user first calls them, which will happen after the `__init__` of the dataset which computes the `Standardizer` and adds it to the transforms..So it should be fine to remove this line now?

ATriantafyllopoulos:
Had to adapt the tests for AIBO to get it working with the new interface, but now it should be ready for review (hopefully for the last time)

autrainer/training/training.py
```diff
@@ -246,7 +246,7 @@ def __init__(
         _preprocess_pipe = SmartCompose([])
         _file_handler = self.data.file_handler
         _features_subdir = cfg.dataset.get("features_subdir", "default")
-        if _features_subdir != "default":
+        if not isinstance(self.data.file_handler, AudioFileHandler):
```
ATriantafyllopoulos, line 249: I do not like the current check either, but it was necessary as now we allow other subdirs and not only default.

autrainer/datasets/aibo.py
ramppdev: Yes still needed due to `train_data = torch.cat([x for x, *_ in self.train_dataset])` instantiating the dataset, e.g. after that we need to manually pass the updated transform again. For dev and test its fine since these are not called yet.

autrainer/training/training.py
ramppdev: Yeah this is a bit tricky, but since we are audio first it should be fine, but i also dont have a "long term" solution at the moment.

ramppdev:
Some tests are failing as `preprocessing/default` should be skipped, just tag me once all are passing and i will review it :)

autrainer/training/training.py
ATriantafyllopoulos: It actually breaks things as it won't let `ToyTabular-C` etc. run through as it then looks for `preprocessing/default`...

ATriantafyllopoulos: I settled for the even uglier:

```python
if (
            not isinstance(self.data.file_handler, AudioFileHandler)
            and _features_subdir != "default" 
        ):
```

ATriantafyllopoulos:
> Some tests are failing as `preprocessing/default` should be skipped, just tag me once all are passing and i will review it :)

Ready!

ramppdev (commented):
Perfect! Only super small comments from my side that are technically not real issues :) == Already accepted after `--num-workers` defaults to 0.

autrainer/datasets/abstract_dataset.py
```diff
@@ -52,6 +53,9 @@ def __init__(
         Args:
             path: Root path to the dataset.
             features_subdir: Subdirectory containing the features.
+                If `None`, defaults to audio subdirectory,
+                which is `default` for the standard format,
+                but can be overridden in the dataset specification.
```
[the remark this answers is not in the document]
ATriantafyllopoulos: Oh absolutely, be nitpicky :) If it's not for urgent bugfixing, we can afford it, I will add it

autrainer/core/scripts/preprocess_script.py
```diff
@@ -24,6 +25,90 @@ class PreprocessArgs(PreprocessArgs):
...
+    )
+    for d, n in (
+        (data.train_dataset, "train"),
+        (data.dev_dataset, "dev"),
+        (data.test_dataset, "test"),
+    ):
+        loader = torch.utils.data.DataLoader(
+            dataset=d,
+            shuffle=False,
+            num_workers=num_workers,
```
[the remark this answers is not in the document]
ATriantafyllopoulos: All fixed in 721647e

autrainer/core/scripts/preprocess_script.py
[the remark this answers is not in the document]
ATriantafyllopoulos: Added a warning in the docs in 6bf34c5ec3f79b5b1e2cbdc12d64963d76facd2e

autrainer/datasets/abstract_dataset.py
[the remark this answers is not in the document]
ATriantafyllopoulos: Fixed in dd04b156703f85ed9a52976cfcd8eb53f80bb669

ATriantafyllopoulos:
If tests past, I will have a final look and merge soon to move on with the other PRs.

autrainer/core/scripts/preprocess_script.py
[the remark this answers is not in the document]
ramppdev: In the CLI wrapper function it is mixed up: `num_workers: int = 1` is still set to 1 by default (should be zero) and the docs for `num_workers` should default to zero instead of the `update_frequency`

PR #61, opened 2024-11-21
Overhaul data pipeline using structs

### LINKED ISSUE #41: Switch to dataclass for data flow
Currently, we rely on tuples for our data flow, i.e. [this is what our datasets return](https://github.com/autrainer/autrainer/blob/d1ca7efbb2a42961ebf4313ef8546e8b1856678a/autrainer/datasets/utils/dataset_wrapper.py#L88) and this is what everything else expects, e.g. [the training loop](https://github.com/autrainer/autrainer/blob/d1ca7efbb2a42961ebf4313ef8546e8b1856678a/autrainer/training/training.py#L454)

I think it's more modular to switch to `@dataclass` objects. This allows us to extend our framework more flexibly, especially regarding models. For example, I am thinking about introducing new tasks that might require additional inputs to the model that come from the dataset. I think the easiest way is to have a base `@dataclass` object that I would then override to add additional parameters that need to be carried around.

Potentially, we need to use this for model outputs too, and use it as the interface between models/losses/transforms.

ATriantafyllopoulos:
Closes #41 

@ramppdev I'd like your opinion here, but no need to do a full-blown review if you don't have the time. I switched to a `@dataclass` `Data` object throughout the code, but needed two major changes.

`Data` has the following mandatory fields: `features`, `labels` and `index`.

The major changes are:

1) Introduce a `ModelWrapper` class with the following logic

```python
class ModelWrapper:
     def __init__(self, model: AbstactModel):
            self.model = model
     def __call__(self, x: Data):
            return self.model(x.features)
```

This is needed because `torchinfo.summary` did not allow me to pass non-tensor inputs to the model. As a result of this change, I had to switch to `self.model.model` throughout `training.py` to save/load states, call `.train()/.eval()` and map to `device`. But this should be fine as the user is unaware of the changes

2) Adapt the default `collate_fn` to handle the new struct, and also change the image aumgentation transforms to handle this composite collating like before. Is there any other place besides the `Mixup` transforms where I need to account for this?

The functionality for this feature change is there. I need to add docs and tests. But I will first merge #54 before proceeding


autrainer/transforms/smart_compose.py
```diff
@@ -72,13 +72,16 @@ def _sort(self):
         """Sort the transforms by their order attribute if present."""
         self.transforms.sort(key=lambda x: getattr(x, "order", 0))
 
-    def get_collate_fn(self, data: "AbstractDataset") -> Optional[Callable]:
+    def get_collate_fn(
+        self, data: "AbstractDataset", default: Callable
```
ramppdev, line 76: Why is it beneficial to have the default collate function as argument here (or not provide the default directly)?
The intention was that augmentations could specify [`get_collate_fn`](https://autrainer.github.io/autrainer/modules/augmentations.html#augmentations-with-collate-functions) and do essentially whatever they want with the data. In some cases it makes sense to just use `torch.utils.data.default_collate`, however we should then just document that we have a custom one instead of providing it always?

ramppdev:
Instead of `labels` i would call it `target` such that the meaning is more generic (thinking towards regression and for the future).

1. Sounds like a valid solution, however this would complicate interacting with the trainer through callbacks etc. Why not just temporarily wrap the model for the summary print or switch out the forward function in a try finally block?
2. Yes the `CutMixUp` is the only augmentation that currently has a collate function.

Other than that the direction looks great and this will make autrainer wayy more extensible :)

ATriantafyllopoulos:
> 1. Sounds like a valid solution, however this would complicate interacting with the trainer through callbacks etc. Why not just temporarily wrap the model for the summary print or switch out the forward function in a try finally block?

I try this approach in 699f2f6, works, but makes the `model_summary.txt` look as follows:

```
========================================================================================================================
Layer (type (var_name))                  Input Shape          Output Shape         Param #              Trainable
========================================================================================================================
ModelWrapper (ModelWrapper)              [1, 64]              [1, 10]              --                   True
├─FFNN (model)                           --                   [1, 10]              --                   True
│    └─Linear (Linear0)                  [1, 64]              [1, 64]              4,160                True
│    └─ReLU (ReLU0)                      [1, 64]              [1, 64]              --                   --
│    └─Dropout (Dropout0)                [1, 64]              [1, 64]              --                   --
│    └─Linear (Linear1)                  [1, 64]              [1, 10]              650                  True
========================================================================================================================
Total params: 4,810
Trainable params: 4,810
Non-trainable params: 0
Total mult-adds (M): 0.00
========================================================================================================================
Input size (MB): 0.00
Forward/backward pass size (MB): 0.00
Params size (MB): 0.02
Estimated Total Size (MB): 0.02
========================================================================================================================
```
I guess this is something we can live with?

ramppdev:
> > 1. Sounds like a valid solution, however this would complicate interacting with the trainer through callbacks etc. Why not just temporarily wrap the model for the summary print or switch out the forward function in a try finally block?
> 
> I try this approach in 699f2f6, works, but makes the `model_summary.txt` look as follows:
> 
> ```
> ========================================================================================================================
> Layer (type (var_name))                  Input Shape          Output Shape         Param #              Trainable
> ========================================================================================================================
> ModelWrapper (ModelWrapper)              [1, 64]              [1, 10]              --                   True
> ├─FFNN (model)                           --                   [1, 10]              --                   True
> │    └─Linear (Linear0)                  [1, 64]              [1, 64]              4,160                True
> │    └─ReLU (ReLU0)                      [1, 64]              [1, 64]              --                   --
> │    └─Dropout (Dropout0)                [1, 64]              [1, 64]              --                   --
> │    └─Linear (Linear1)                  [1, 64]              [1, 10]              650                  True
> ========================================================================================================================
> Total params: 4,810
> Trainable params: 4,810
> Non-trainable params: 0
> Total mult-adds (M): 0.00
> ========================================================================================================================
> Input size (MB): 0.00
> Forward/backward pass size (MB): 0.00
> Params size (MB): 0.02
> Estimated Total Size (MB): 0.02
> ========================================================================================================================
> ```
> 
> I guess this is something we can live with?

Yes for me this would be totally fine. If we really dont like it we could later temporarily override the forward method with a custom one, and in a try block get the summary, and in the final block switch the forward function back to the original one. Should then be safe and not have the `ModelWrapper` in the print. But at the same time weird code that is kinda unnecessary.

ATriantafyllopoulos:
@ramppdev tests should be passing now, or soon :)

Would be great if you can do a more in-depth review whenever you find the time. 

I realized while writing this that the `dataclass` would not work nicely for `serving`, so I added an `if-clause` to allow `torch.Tensor` inputs, which I assume to be the features and propagate them directly.

Other than that, please have a look at the issue with `collate_fn`, I am not yet entirely certain how much the tests actually cover in terms of functionality and correctness for the mixup transforms..

autrainer/datasets/utils/dataloader.py
```diff
@@ -0,0 +1,24 @@
+from collections.abc import Mapping
+from typing import List
+
+import numpy as np
+import torch
+
+from .data_struct import Data
+
+
+def default_data_collator(features: List[Data]) -> Data:
```
ramppdev, line 10: I think a docstring would be great here especially covering what happens with tensors, numpy arrays, or e.g. ints.

ramppdev:
> Other than that, please have a look at the issue with `collate_fn`, I am not yet entirely certain how much the tests actually cover in terms of functionality and correctness for the mixup transforms..

Generally speaking for transforms and augmentations the tests only cover  basic in -> out sanity checks, they do not really tests functionality as this is quite difficult for most of them.

autrainer/training/training.py
```diff
@@ -778,19 +780,20 @@ def _evaluate(
...
                     self.criterion(
                         probabilities_fn(output),
-                        target.to(self.DEVICE),
+                        data.target,
                     )
                     .cpu()
                     .item()
                 )
-                tracker.update(output, target, sample_idx)
+                tracker.update(output, data.target.cpu(), data.index.cpu())
```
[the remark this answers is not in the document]
ATriantafyllopoulos: No, it's just for safety, I guess it will be ignored..

autrainer/datasets/utils/data_struct.py
```diff
@@ -0,0 +1,16 @@
...
+@dataclass
+class Data:
+    features: torch.Tensor
+    target: Union[int, float, List[int], List[float]]
+    index: int
+
+    def to(self, device):
+        self.features.to(device)
+        self.target.to(device)
+        self.index.to(device)
```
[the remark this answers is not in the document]
ATriantafyllopoulos: Hmm, good point, I wanted to move all to GPU by default, but I can skip that..

autrainer/models/abstract_model.py
```diff
@@ -14,8 +17,7 @@ def __init__(self, output_dim: int) -> None:
         super().__init__()
         self.output_dim = output_dim
 
-    @abstractmethod
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def forward(self, x: Union[Data, torch.Tensor]) -> torch.Tensor:
```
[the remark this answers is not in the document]
ATriantafyllopoulos: Stupid me, I realize I don't need the `_forward` class since I can call `super().forward()`, I'll push an update shortly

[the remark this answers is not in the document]
ATriantafyllopoulos: See 4e606be5a167f8f1f673ed9ee6e104ecca11c91f

ATriantafyllopoulos:
Reply to https://github.com/autrainer/autrainer/pull/61#discussion_r1873969785

For some reason I can't respond inline..anyway...

I am thinking a bit ahead, but new datasets might need to create their custom `collate_fn` if they start introducing more complicated logic. Allowing that option now -- even though it is not needed right away -- seems more "future-proof" to me than hardcoding it

autrainer/models/abstract_model.py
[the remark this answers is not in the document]
ramppdev: Okay i dont get how this will work can you explain please? The `AbstractModel` now has a super call?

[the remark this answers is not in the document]
ATriantafyllopoulos: Exactly. It will call `super().forward(...)`, which begins to resolve right to left and stops at the call to `torch.nn.Module.forward(...)` since the other classes don't have a `forward(...)` method (if they had, that would be a bummer).

So then `AbstractModel` implements the parsing of the data (if it's `Data`, then pass `Data.features`, if it's `torch.Tensor`, then pass the raw input), and leaves the "actual" `forward(...)` implementation to the derived class.

---
P.S.

Now that I think about it with a bit more clear head, I need to double-check the inheritance resolution to make sure we don't end up in some sort of weird infinite recurrence loop and/or break backprop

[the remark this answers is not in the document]
ATriantafyllopoulos: Forget what I said above, I am obviously wrong, I'll try another solution in a bit...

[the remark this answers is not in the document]
ATriantafyllopoulos: Then, to contradict myself again, the whole idea behind this is to have an easy way to override the parsing for custom models. This would speak in favor of a `_parse_data` method that can be overriden by the user in their local code, and thus the current implementation. 

So to summarize the three options as I see them:

1. The user writes a `forward(self, torch.Tensor)` function as normal. `autrainer` wraps it in a `ModelWrapper` under-the-hood to take care of the unwrapping of data. The benefit is that the user doesn't have to care at all about this functionality. However, this makes the whole PR "too much ado about nothing" as the whole point from the start was to make this configurable and allow models to benefit from added entries to the `Data` struct. I would thus drop this option.

2. The user writes a `forward(self, Data)` function. They have to call `AbstractModel._parse_data(self, Data)` or else their model will crash (not sure how we can make this error message more informative). They also have the option to override `AbstractModel._parse_data(self, Data)` and integrate custom logic to utilize more fields of an extended `Data` struct.

3. Similarly, we can force the user to implement a private `_function(self, Tensor)` method and implement `AbstractModel.forward(self, Union[Data, Tensor])` to take care of the unwrapping. The downside of that, is that for custom logic that unpacks more stuff out of `Data`, the user would additionally need to reimplement the `forward(...)` function.

So, in summary, I think I lean towards 2. This is also the solution currently implemented in this branch. This solution sacrifice ease-of-use for extended functionality. I think it's worth it as I will "unlock" a lot more capabilities.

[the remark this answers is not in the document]
ramppdev: Why do we not enforce a strict interface `forward(self, data: Data)` all the time? Since we get this from the Dataset, this feels like it makes the most sense to me if we want to pass full control to the user and let them do whatever they want with the inputs.
E.g. inherit from `Data` (provide lets say an abstract implementation of the bare minimum fields that we need internally), and then they could add their dataclasses on both the dataset and model side?

I feel like we loose the main benefit of moving to dataclasses when we still pass only the Tensor to the model.

ramppdev:
> Reply to [#61 (comment)](https://github.com/autrainer/autrainer/pull/61#discussion_r1873969785)
> 
> For some reason I can't respond inline..anyway...
> 
> I am thinking a bit ahead, but new datasets might need to create their custom `collate_fn` if they start introducing more complicated logic. Allowing that option now -- even though it is not needed right away -- seems more "future-proof" to me than hardcoding it

Ok yes this is a very good idea then :)

autrainer/models/abstract_model.py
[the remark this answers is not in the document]
ATriantafyllopoulos: > Why do we not enforce a strict interface forward(self, data: Data) all the time?

Right now we do, actually, and provide the utility method `self._parse_data()` to handle the default "unpacking" of the struct. But the problem is that `python` does not "enforce" anything, so users coming from native `PyTorch` will have to adapt to the new way of doing things (which is fine, it's what `transformers` have done and it's been very successful there).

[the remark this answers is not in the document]
ramppdev: While your IDE most likely will help you and generate the proper type annotations when you implement the abstract forward method, i definitely see your concerns and having `def forward(self, x: Union[Data, Tensor])->...` is non standard and inconvenient.

If i remember correctly then transformers also somewhat adhere to the PyTorch API and destructure their dataclasses on input?

This would also be a solution to the problem, such that we DO enforce some `standard` structure of the Data struct (+optionally also the forward function arguments).
We could check on the caller side which arguments of the current Data struct "fit in" the forward method and pass them directly as tensors, and enforce that the first one always has to the the features / inputs tensor (this would solve all our current issues). We even could go so far to not care if the first argument is named x, inputs, features, etc. and just always pass the features tensor this way. This is somewhat similar to the transformers implementation. Most forward methods have the other arguments as optional as well here (which we would need as well for inference at the moment).

Lets say a user defines a dataset that additionally returns a `segmentation_mask` attribute for a custom Data struct, we could similarly destructure it and pass it to the appropriate keyword argument.

Down the line, once we implement dataclasses support for transforms and augmentations, it could be the case that a transformer feature extractor transform returns an additional `attention_mask` we should pass to the model (e.g., creates a new Data struct with that field) and then we would handle it the same way we did above.

What do you think about this approach? Are there major drawbacks that i have missed?

autrainer/datasets/utils/data_struct.py
ramppdev, line 1: Something to keep in mind here is that `Data` is an "item-centric" view on the data here, i.e. represents a single data point returned from a dataset.
The `to(...)` method however is only valid when we have a "batch-centric" view on the data, i.e. the representation returned from the loader.
Then the `to` method is valid, but the types are invalid again.
It would probably be best to:
- create an abstract item-centric and batch-centric implementation, implement `to` only for the batched one
- ensure that our required fields (features, target, index) are always present
- transform from the item-centric classes to a single batch-centric class in the collation function

ATriantafyllopoulos:
5a42dac added a tutorial that showcases the use of this more advanced pipeline

ATriantafyllopoulos:
d8656ad added the docs

ATriantafyllopoulos:
@ramppdev please have another good look at the PR :) This is the home-run for merging, I hope

docs/source/modules/dataset_utils.rst
```diff
@@ -1,9 +1,67 @@
...
+needed by :class:`~autrainer.models.AbstractModel`
+objects that are compatible
+with a particular dataset,
+as well as attributes
+needed by `autrainer`'s utilities,
+such as the *index* corresponding to each *instance*
+(where we assume that each :class:`~autrainer.datasets.AbstractDataset`
+is an ordered set of instances).
+
+.. autoclass:: autrainer.datasets.utils.DataItem
```
ramppdev, line 27: Since we have the `AbstractDataBatch` below, should we also add the `AbstractDataItem` here?
Also, in the above description it would be nice to make it clear that a dataset should return such a `AbstractDataItem`.

We could also add this to the normal dataset description and link here which i would actually prefer.

docs/source/modules/dataset_utils.rst
```diff
@@ -1,9 +1,67 @@
...
+   :members:
+
+
+Additionally,
+we provide classes
+that hold *batches* of individual instances,
+and provides an implementation
+of the `collate_fn`
```
ramppdev, line 35: These should be ```... :meth:`collate_fn` ...``` and ```... :class:`torch.utils.data.DataLoader` ...```

docs/source/modules/dataset_utils.rst
```diff
@@ -1,9 +1,67 @@
...
+   :attr:`~autrainer.datasets.utils.AbstractDataItem.target`,
+   and :attr:`~autrainer.datasets.utils.AbstractDataItem.index`
+   are the three **reserved** attributes
+   that every object derived from 
+   :class:`~autrainer.datasets.utils.AbstractDataItem`
+   **must** include.
+   Moreover,
+   every object derived from
+   :class:`~autrainer.models.AbstractModel`
+   **must** include :attr:`~autrainer.datasets.utils.AbstractDataItem.features`
```
ramppdev, line 59: We should link this in the model docs as well.

autrainer/models/abstract_model.py
```diff
@@ -15,23 +18,32 @@ def __init__(self, output_dim: int) -> None:
         self.output_dim = output_dim
 
     @abstractmethod
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
ramppdev, line 18: The `forward` method went missing :D optionally, we could change the input to `embeddings` to features as well.

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
 of a :ref:`dataset <datasets>` implementation.
 
+.. _tut_advanced:
+
+Advanced Data Pipelines
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To create data and model pipelines that go beyond the standard :class:`~autrainer.dataset.DataItem` convention of using
+only :attr:`DataItem.features` as input to the model, first create a new :class:`~autrainer.dataset.DataItem` struct
+decorated with :class:`dataclasses.@dataclass`:
```
ramppdev, line 289: `~dataclasses.@dataclass` (or without the @)

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
 The :ref:`target transforms <dataset_target_transforms>` are specified in the :attr:`~autrainer.datasets.AbstractDataset.target_transform` property
 of a :ref:`dataset <datasets>` implementation.
 
+.. _tut_advanced:
+
+Advanced Data Pipelines
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To create data and model pipelines that go beyond the standard :class:`~autrainer.dataset.DataItem` convention of using
```
ramppdev, line 287: The namespaces are missing `utils` here: `~autrainer.datasets.utils.DataItem` (in the whole file).

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
+only :attr:`DataItem.features` as input to the model, first create a new :class:`~autrainer.dataset.DataItem` struct
+decorated with :class:`dataclasses.@dataclass`:
+
+.. literalinclude:: ../examples/tutorials/multi_branch_model.py
+   :language: python
+   :caption: data_struct.py
+   :lines: 10-15
```
ramppdev, line 294: All the lines in this file are a bit offset, i think they need to be shifted 3-4 lines down.

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
+
+and override :class:`~autrainer.dataset.AbstractDataBatch`:
+
+.. literalinclude:: ../examples/tutorials/multi_branch_model.py
+   :language: python
+   :caption: data_batch.py
+   :lines: 28-46
+
+Following that, inherit :class:`~autrainer.datasets.dataset_wrapper.DatasetWrapper`
```
ramppdev, line 303: "...inherit *from*..." (same for all "inherit from" instances in this file) and namepace should be: `autrainer.datasets.utils.DatasetWrapper` (it is not linked atm; same below).

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
+to create a :class:`torch.utils.data.Dataset`
+that iterates over your data 
+and returns your custom :class:`~autrainer.dataset.AbstractDataBatch`
+(here we simply replicate :attr:`~autrainer.datasets.ToyDatasetWrapper.features`
```
ramppdev, line 307: `autrainer.datasets.ToyDatasetWrapper.features` does not exist but can even be omitted to just features i guess.

docs/source/examples/tutorials/MultiBranchModel.yaml
```diff
@@ -0,0 +1,7 @@
+id: ToyMultiBranchModel
+_target_: model.ToyMultiBranchModel
```
ramppdev, line 2: We import this with: `.. literalinclude:: ../examples/tutorials/multi_branch_model.py`, so i would recommend to change the target to: `_target_: multi_branch_model.ToyMultiBranchModel`.

docs/source/examples/tutorials/MultiBranchData-C.yaml
```diff
@@ -0,0 +1,20 @@
+
+id: ToyMultiBranch-C
+_target_: dataset.ToyMultiBranchData
```
ramppdev, line 3: I would change the target to: `_target_: multi_branch_dataset.ToyMultiBranchData` (see below).

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
+as our auxiliary features):
+
+.. literalinclude:: ../examples/tutorials/multi_branch_model.py
```
ramppdev, line 310: For readability, i would change the captions here to `multi_branch_dataset.py` for the data related stuff and `multi_branch_model.py` for the model, this way it aligns with the imports in the configs and the user could just copy paste this.

docs/source/examples/tutorials/multi_branch_model.py
```diff
@@ -0,0 +1,94 @@
...
+import torch
+
+from autrainer.datasets.toy_dataset import ToyDataset, ToyDatasetWrapper
+from autrainer.datasets.utils.data_struct import AbstractDataBatch
+from autrainer.models import AbstractModel
+from autrainer.transforms import SmartCompose
+
+
+@dataclass
+class DataItemMultiBranch:
```
ramppdev, line 13: Shouldnt this inherit from `AbstractDataItem`?

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
+
+.. literalinclude:: ../examples/tutorials/MultiBranchData-C.yaml
+   :language: yaml
+   :caption: conf/data/MultiBranchData-C.yaml
+   :linenos:
+
+Finally, you must inherit from :class:`~autrainer.datasets.AbstractModel` 
+and create a model **with a matching signature**,
+i.e. one which includes both :attr:`features`
+and the additional variables
```
ramppdev, line 336:
```
Finally, you must inherit from :class:`~autrainer.datasets.AbstractModel` and create a model **with a matching signature** in its forward pass to access the :attr:`meta` parameter:
```

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
+besides :attr:`target` and :attr:`index`
+in its :meth:`~autrainer.datasets.AbstractModel.forward` signature:
+
+.. literalinclude:: ../examples/tutorials/multi_branch_model.py
+   :language: python
+   :caption: model.py
+   :lines: 69-86
+
+
+This model will be configured with hydra:
```
ramppdev, line 346:
```
Next, create a :attr:`ToyMultiBranchModel.yaml` configuration file for the model in the :attr:`conf/model/` directory:
```
caption should be `conf/model/ToyMultiBranchModel.yaml`

docs/source/usage/tutorials.rst
```diff
@@ -278,6 +279,77 @@ For example, the following target transform logarithmically encodes and decodes
...
+to create a dataset
+that instantiates your :class:`~autrainer.datasets.dataset_wrapper.DatasetWrapper`:
+
+
+.. literalinclude:: ../examples/tutorials/multi_branch_model.py
+   :language: python
+   :caption: dataset.py
+   :lines: 49-67
+
+This is the dataset you will configure
```
ramppdev, line 325:
```
Next, create a :attr:`ToyMultiBranch-C.yaml` configuration file for the dataset in the :attr:`conf/dataset/` directory:
```
caption should be `conf/dataset/ToyMultiBranch-C.yaml`

ramppdev (approved):
Code is perfect!
Sorry i noticed half way through that i could just open a PR for the docs but didnt think i would be so extremely nitpicky :D

docs/source/examples/optimizer_custom_step.py
```diff
@@ -2,24 +2,30 @@
...
     ) -> Tuple[torch.Tensor, torch.Tensor]:
         """Custom step function for the optimizer.
 
         Args:
-            model: Model to be optimized.
-            data: Batched input data.
-            target: Batched target data.
+            model: The model to train.
+            data: The data batch containing features, target, and potentially
+                additional fields. The data batch is expected to be on the same
```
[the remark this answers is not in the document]
ATriantafyllopoulos: So should I change something here?

docs/source/modules/dataset_utils.rst
ATriantafyllopoulos: The reason I skipped it is it has no docs, should I add some?

docs/source/examples/tutorials/MultiBranchData-C.yaml
ramppdev, line 1: would be best if the name of the file is also aligned with the id to find it easier

docs/source/examples/tutorials/MultiBranchModel.yaml
ramppdev, line 1: same with id and name here

docs/source/modules/dataset_utils.rst
ramppdev: Yeah that would be great, i accidentally missed that then.

ramppdev (approved):
Looks good so far, i guess once these comments are integrated we are ready to merge :)

docs/source/examples/optimizer_custom_step.py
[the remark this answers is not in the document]
ramppdev: Probably you can drop the "expected to be".

ATriantafyllopoulos:
> Looks good so far, i guess once these comments are integrated we are ready to merge :)

Maybe you can resolve the ones that are already addressed with the newest changes? Or should I?

ramppdev:
> > Looks good so far, i guess once these comments are integrated we are ready to merge :)
> 
> Maybe you can resolve the ones that are already addressed with the newest changes? Or should I?

Probably makes most sense if you go over them a final time and if all are resolved / ignored (some totally optional) go ahead an merge :)

Issue #64, opened 2024-11-27
Self-supervised training


ATriantafyllopoulos:
Following up on #61, I plan to extend `autrainer` with a few standard "recipes" for self-supervised training. I will do this by introducing new datasets/tasks, optionally also models. I will begin with simple auto-encoders for a start, and then slowly expand to denoisng AEs, masking, and others..

ramppdev:
Sounds like a very good idea, this might be the time that we modularize the `training.py` "script" to easier handle more complex training setups.

PR #65, opened 2024-11-27
Added simple dataset overview to bookkeeping

millinma:
This PR would add some simple dataset overview of info that seem IMO relevant info for any given task. For classification and ML classification it determines the number of classes, as well as size of max/min class + data imbalance. For regression it calculates mean, std min and max for the given label. Total number of samples is always calculated. In this version it also shouldn't break with new task types like SSL #64 or multi-label regression, but could/should potentially be extended for these.

For the moment, I did not know how to best deal with the validation set. Sometimes there is no dataset available, sometimes it is just the test dataset, so the default is to just save it for train and test. 

Also, these were just the info points that come "for free", what might be interesting information to extend to would be e.g., total audio length for audio files or image sizes for image data etc., but for that we might need to load each data point once already in the bookkeeping.

### DIFF
--- a/autrainer/core/utils/bookkeeping.py
+++ b/autrainer/core/utils/bookkeeping.py
@@ -101,6 +101,76 @@ class Bookkeeping:
             exist_ok=True,
         )
 
+    def save_dataset_summary(
+        self,
+        data: "AbstractDataset",
+        filename: str,
+        val_exists: bool = False,
+    ) -> None:
+        """Save some overview of the dataset to a file.
+
+        Args:
+            data: dataset object
+            filename: Name of the file to save the overview to.
+        """
+        if val_exists:
+            split_datasets = [data.train_dataset, data.val_dataset, data.test_dataset]
+            index = ["Train", "Development", "Test", "Total"]
+        else:
+            split_datasets = [data.train_dataset, data.test_dataset]
+            index = ["Train", "Test", "Total"]
+        n_samples = []
+        if isinstance(data, BaseClassificationDataset) or isinstance(data, BaseMLClassificationDataset):
+            n_classes = []
+            n_majority_class = []
+            n_minority_class = []
+            class_imbalance = []
+        elif isinstance(data, BaseRegressionDataset):
+            mean_label = []
+            std_label = []
+            max_value = []
+            min_value = []
+        
+        data_dfs = [dataset.df for dataset in split_datasets]
+        aggregated_data_df = pd.concat(data_dfs, axis=0, ignore_index=True)
+        data_dfs.append(aggregated_data_df)
+        target_column = split_datasets[0].target_column
+        for data_df in data_dfs:
+            n_samples.append(data_df.shape[0])
+            if isinstance(data, BaseClassificationDataset):
+                label_counts = data_df[target_column].value_counts()
+                n_classes.append(label_counts.count())
+                n_majority_class.append(label_counts.max())
+                n_minority_class.append(label_counts.min())
+                class_imbalance.append(label_counts.max() / label_counts.min())
+            elif isinstance(data, BaseMLClassificationDataset):
+                label_counts = data_df[target_column].sum(axis=0)
+                n_classes.append(len(target_column))
+                n_majority_class.append(label_counts.max())
+                n_minority_class.append(label_counts.min())
+                class_imbalance.append(label_counts.max() / label_counts.min())
+            elif isinstance(data, BaseRegressionDataset):
+                label_values = data_df[target_column].values
+                mean_label.append(label_values.mean())
+                std_label.append(label_values.std())
+                max_value.append(label_values.max())
+                min_value.append(label_values.min())
+        if isinstance(data, BaseClassificationDataset) or isinstance(data, BaseMLClassificationDataset):
+            data_overview_df = pd.DataFrame(
+                {"n_samples": n_samples, "n_classes": n_classes, "n_majority_class": n_majority_class, 
+                "n_minority_class": n_minority_class, "imbalance_ratio": class_imbalance},
+                index=index) 
+        elif isinstance(data, BaseRegressionDataset):
+            data_overview_df = pd.DataFrame(
+                {"n_samples": n_samples, "mean_label": mean_label, "std_label": std_label, 
+                "max_value": max_value, "min_value": min_value},
+                index=index)
+        else:
+            data_overview_df = pd.DataFrame(
+                {"n_samples": n_samples},
+                index=index)
+        data_overview_df.to_csv(os.path.join(self.output_directory, filename))
+
     def save_model_summary(
         self,
         model: torch.nn.Module,
--- a/autrainer/training/training.py
+++ b/autrainer/training/training.py
@@ -154,6 +154,9 @@ class ModularTaskTrainer:
                 state_dict,
                 skip_last_layer,
             )
+        self.bookkeeping.save_dataset_summary(
+            self.data, "dataset_summary.csv"
+        )
         self.bookkeeping.save_model_summary(
             self.model, self.train_dataset, "model_summary.txt"
         )


autrainer/core/utils/bookkeeping.py
```diff
@@ -101,6 +101,76 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
             exist_ok=True,
         )
 
+    def save_dataset_summary(
+        self,
+        data: "AbstractDataset",
+        filename: str,
+        val_exists: bool = False,
```
ramppdev, line 108: How do we get this? I think `val_exists=True` is unreachable from the current setup in the `training.py`.

autrainer/core/utils/bookkeeping.py
```diff
@@ -101,6 +101,76 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
...
+        else:
+            split_datasets = [data.train_dataset, data.test_dataset]
+            index = ["Train", "Test", "Total"]
+        n_samples = []
+        if isinstance(data, BaseClassificationDataset) or isinstance(data, BaseMLClassificationDataset):
+            n_classes = []
+            n_majority_class = []
+            n_minority_class = []
+            class_imbalance = []
+        elif isinstance(data, BaseRegressionDataset):
```
ramppdev, line 128: We also support multi-target regression now (`BaseMTRegressionDataset`).

autrainer/core/utils/bookkeeping.py
```diff
@@ -101,6 +101,76 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
...
+            mean_label = []
+            std_label = []
+            max_value = []
+            min_value = []
+        
+        data_dfs = [dataset.df for dataset in split_datasets]
+        aggregated_data_df = pd.concat(data_dfs, axis=0, ignore_index=True)
+        data_dfs.append(aggregated_data_df)
+        target_column = split_datasets[0].target_column
+        for data_df in data_dfs:
```
ramppdev, line 138: The following cases may be simplified if we add an abstract `summary` method (or similar) to the `AbstactDataset` that is implemented by all base datasets. This way we can keep the functionality of the `Bookkeeping` class to just exporting / saving to disk here, and if needed, custom datasets could override the summary. What are your thoughts?

ramppdev (commented):
Good idea, this is very strong for reporting purposes.

autrainer/core/utils/bookkeeping.py
millinma: Yes, I know. I have it here as an option, but we would need some rule on how to determine whether to do it for validation or not. I think the 2 possibilities where we do not want it is if 1) no validation exists (obviously) and 2) if validation is equal to test. The latter I am not sure how to determine easily, do you have an idea @ramppdev?

autrainer/core/utils/bookkeeping.py
millinma: Ah i missed that this is already supported, anyways, there maybe it would make sense to do the same as for single target regression but for all targets separately?

autrainer/core/utils/bookkeeping.py
millinma: Yea, that might be a good idea. But shall we add it to the `AbstactDataset` or rather individually to all the task-specific abstract datasets like `BaseClassificationDataset`?

autrainer/core/utils/bookkeeping.py
ramppdev: At the moment, a validation dataset always exists, but could be equal to the test dataset. So the easiest way is to just always output it.
If we choose to move the logic to the datasets we could individually handle it there.

autrainer/core/utils/bookkeeping.py
ramppdev: Yep this should be good.

autrainer/core/utils/bookkeeping.py
ramppdev: We should add an abstract method to the abstract dataset and implement the defaults you have here in the base dataset classes. If specific features are needed then, a dataset could just override the method again.

autrainer/core/utils/bookkeeping.py
```diff
@@ -101,6 +106,96 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
             exist_ok=True,
         )
 
+    def save_dataset_summary(
```
ATriantafyllopoulos, line 109: I tend to think that this should not be a "mandatory" part of the `bookkeeping` that's executed during `autrainer train`, but rather a separate script that can be optionally called by the user with a new command (e.g. `autrainer datasummary`). 

There's a few arguments in favor of this:

a) The feature should be easy to change by the user locally, in case they want to extend/override some of the functionality. This makes me think it should be implemented like a standalone class (more on that below), because in the place where it is now, the user cannot touch it

b) While not critical, it does add some extra overhead to 

c) Conceptually (I think this has been raised in an issue by @ramppdev that this is unique per `dataset` configuration in the `config.yaml` but *not* per `dataset/model` pair, i.e. per "training run". So there is also conceptually not needed to run it every time.

---

To immediately take the other position, I also see the benefit of packaging this along with every training run so when we want to publish (or store) a particular run, we immediately get the data for free.

I am open for this approach as well, but I think in any case we should achieve the goal of modularity for the end-user

autrainer/core/utils/bookkeeping.py
```diff
@@ -101,6 +106,96 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
...
+        self,
+        data: "AbstractDataset",
+        filename: str,
+        val_exists: bool = False,
+    ) -> None:
+        """Save some overview of the dataset to a file.
```
ATriantafyllopoulos, line 115: "some" is not really a good description xD

autrainer/core/utils/bookkeeping.py
ATriantafyllopoulos: With the new format #54, it should be more or less always there for each dataset (even if it is identical)

autrainer/core/utils/bookkeeping.py
ATriantafyllopoulos: I am also in favor of this approach, and this covers the "modularity" concern I also express in my review

autrainer/core/utils/bookkeeping.py
```diff
@@ -101,6 +106,96 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
...
+                {
+                    "n_samples": n_samples,
+                    "n_classes": n_classes,
+                    "n_majority_class": n_majority_class,
+                    "n_minority_class": n_minority_class,
+                    "imbalance_ratio": class_imbalance,
+                },
+                index=index,
+            )
+        elif isinstance(data, BaseRegressionDataset):
```
ATriantafyllopoulos, line 182: Why not default to `df[...].describe()`? This already covers most of the stuff we need

autrainer/core/utils/bookkeeping.py
```diff
@@ -101,6 +106,96 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
...
+            elif isinstance(data, BaseRegressionDataset):
+                label_values = data_df[target_column].values
+                mean_label.append(label_values.mean())
+                std_label.append(label_values.std())
+                max_value.append(label_values.max())
+                min_value.append(label_values.min())
+        if isinstance(data, BaseClassificationDataset) or isinstance(
+            data, BaseMLClassificationDataset
+        ):
+            data_overview_df = pd.DataFrame(
```
ATriantafyllopoulos, line 172: Might be better to show the number of  samples per class by doing `self.df[self.target_column].value_counts()`? Then it's easy to extract the majority/minority staff

ATriantafyllopoulos (changes requested):
I made some more minor comments inline, but I agree with @ramppdev that the functionality should be transferred to `AbstractDataset` and modularized there, so I'll wait first for those changes before reviewing again.

In general, I think it's useful and in the right direction!

Before merging, we will also need to add some tests for this functionality, but it should be easy once moved to the datasets

millinma:
> I made some more minor comments inline, but I agree with @ramppdev that the functionality should be transferred to `AbstractDataset` and modularized there, so I'll wait first for those changes before reviewing again.

Yes, I think actually has to be moved there, even though I think it works for all the standardised datasets we have now, on our custom CIFAR implementation it breaks. So I will go ahead and move it there.

autrainer/core/utils/bookkeeping.py
millinma: > I tend to think that this should not be a "mandatory" part of the `bookkeeping` that's executed during `autrainer train`, but rather a separate script that can be optionally called by the user with a new command (e.g. `autrainer datasummary`).
> 
> There's a few arguments in favor of this:
> 
> a) The feature should be easy to change by the user locally, in case they want to extend/override some of the functionality. This makes me think it should be implemented like a standalone class (more on that below), because in the place where it is now, the user cannot touch it

Do you think this concern is resolved if the user can overwrite the data_overview() function of the dataset?

> 
> b) While not critical, it does add some extra overhead to

Shall we then maybe add an option to enable/disable it? I think runtime-wise it is relatively neglectable compared to the training?

> 
> c) Conceptually (I think this has been raised in an issue by @ramppdev that this is unique per `dataset` configuration in the `config.yaml` but _not_ per `dataset/model` pair, i.e. per "training run". So there is also conceptually not needed to run it every time.

Sure, those points technically make sense to me too, but in our current pipeline I do not see in which step we could include this (and I do not think it is worth its own cli command?). Especially since e.g., validation splits only come into play for the training itself, and not for instance in the fetch command?

ramppdev: > c) Conceptually (I think this has been raised in an issue by @ramppdev that this is unique per `dataset` configuration in the `config.yaml` but _not_ per `dataset/model` pair, i.e. per "training run". So there is also conceptually not needed to run it every time.

For the sake of simplicity and ease of use of these statistics in a downstream task, we could run it for every "training run" instead of once per dataset. I see it as a benefit if this information is always located in the same spot from a run-based perspective.

PR #82, opened 2024-12-21
Add Support for Global Transforms

### LINKED ISSUE #56: Transforms Setup Function for Full Dataset Access
All transforms only receive a single sample during training / dev / test.
Some transforms like the `Standardizer` of the `AIBO` dataset require information about the full dataset, which we currently solve by placing them in the dataset init: https://github.com/autrainer/autrainer/blob/e1ecb6c2a4e98ed6e00abd9d6423117bfdb9abbb/autrainer/datasets/aibo.py#L98-L112

Therefore, add a new method `fit` or `setup` providing all transforms with full access to the dataset after initialization.
E.g., this would allow specifying the `Standardizer` in the dataset configuration instead of the init, and during setup it would read the mean and standard deviation of the training dataset, afterwards applying the same transform to all subsets.

However, this is not fully straight forward as we use `audobject` to export the transformation pipeline during training and these dynamically set values need to be accounted for.

Ref: https://github.com/autrainer/autrainer/pull/54#discussion_r1824290604

ramppdev:
Closes #56 by adding support for transforms that receive globally and dynamically computed parameters.
Transforms that need to capture global statistics (like the mean and std for the `Standardizer` previously hardcoded the `AIBO` dataset) now can use a `autrainer.transforms.GlobalTransform` wrapper (see example below):
- the wrapper calls the `from_global` class method and passes in the dataset subset the transform is currently built for
- by default, the wrapper takes in the position of the underlying transform and allows control of whether to skip possible augmentations in the transformation pipeline or to use the actual correct computed position in the transformation pipeline when resolving the global transform
- the `from_global` class method of the underlying transform should instantiate itself after computing the relevant parameters
- after the underlying transform is built, the wrapper does the ol switcheroo and replaces itself with the actual transform instance such that it can be correctly serialized
- switching the transform in this was has important reasons:
  - `audobject` records the arguments that are passed to a class during instantiation, meaning we cant serialize and deserialize the class later on when the concrete values depend on some internal computation and the dataset; basically we have to do it this way to trick `audobject` into thinking the class was instantiated as usual
  - it generally makes sense to resolve these values and have them "hardcoded" in the resulting yaml files for ease of use

This now allows us to replace the `Standardizer` used in the `AIBO` dataset with our `autrainer.transforms.Normalize` class we use everywhere else.
I have extended it to handle 1-3D input data and with a `from_global` class method to compute the normalization of the current dataset on the fly.
Therefore the AIBO dataset again standardize its training data in the following way:

```yaml
...
transform:
  ...
  train:
    - autrainer.transforms.GlobalTransform:
        transform: autrainer.transforms.Normalize
        order: -100 # defaults to Normalize's order if not specified
```

If there would be a `autrainer.transforms.ScaleRange` transform scaling to the range [0,1] preceding the normalization in the computed transformation order, this IS also accounted, e.g. we dynamically resolve the data by the actual position in the pipeline if desired.

---

However, when implementing i noticed the following bugs in our current implementation:
- we allow [overriding or removing transforms](https://autrainer.github.io/autrainer/modules/transforms.html#id5) by setting them to `null`, this was only the case for each subset, e.g. if a model tries to override / remove a transform specified in the `base` subset, this was not possible. Now i have changed it to globally resolve for each subset (including base) separately.
- due to the override / removal, we convert model and dataset transform lists to sets before instantiating, which only allows one "class" of transforms per pipeline (this is not the case for augmentations and never documented, therefore very unusual for the user to understand / i would even say a bug). This is now changed such that multiple transforms of the same type can be specified
- since we now can have multiple transforms of the same type, overriding / removing a transform is no longer a 1:1 mapping. Therefore i added a `<transform>[@<tag>]` syntax only used for identification of the transforms allowing for greater flexibility (see below). This should allow for the complete control over the pipelines if needed.

```yaml
# SomeDataset.yaml
transform:
  ...
  base:
    - autrainer.transforms.AnyToTensor@base
  train:
    - autrainer.transforms.AnyToTensor@someTag:
        order: 123
    - autrainer.transforms.AnyToTensor

# SomeModel.yaml
transform:
  ...
  train:
    - autrainer.transforms.AnyToTensor@base: null

```

This example would override the `@base` tagged transform in the base transforms ONLY for the training transform pipeline. For dev, and test, it would still be included. All other transforms also would be included in the training pipeline.

I know, at this point we are basically writing our own DSL for setting up autrainer :D so let me know what you think about these changes @ATriantafyllopoulos.

---

Still draft, as the docs are not updated and tests are missing:
- [x] Update documentation
- [x] Update tests


autrainer/transforms/smart_compose.py
```diff
@@ -90,6 +93,25 @@ def get_collate_fn(self, data: "AbstractDataset") -> Optional[Callable]:
...
+        for i, t in enumerate(self.transforms):
+            if isinstance(t, GlobalTransform):
+                self._skip_augmentations = t.skip_augmentations
+                if t.resolve_by_position:
+                    c = SmartCompose(self.transforms[:i])
+                    c._skip_augmentations = self._skip_augmentations
+                    data.transform = c
+                t.setup(data)
+                self._skip_augmentations = False
+                data.transform = self
```
ATriantafyllopoulos, line 113: Ugh, this is very, very opaque. Need to think it through, but this now makes it all very convoluted

ATriantafyllopoulos: I wouldn't let this overwrite `data.transform` but would change the line that calls this 
from: `self.transform.setup(self)`
to 

```python
# init transform using current data
transform.setup(self) 
# overwrite transform after init
self.transform = transform
```

--- 

I would also consider moving it to `_init_dataset`, though I guess this is harder to do?

```python
data = DatasetWrapper(
            path=self.features_path,
            features_subdir=self.features_subdir,
            index_column=self.index_column,
            target_column=self.target_column,
            file_type=self.file_type,
            file_handler=self.file_handler,
            df=df,
            transform=None,
            target_transform=self.target_transform,
        )
transform.setup(data)
return DatasetWrapper(
            path=self.features_path,
            features_subdir=self.features_subdir,
            index_column=self.index_column,
            target_column=self.target_column,
            file_type=self.file_type,
            file_handler=self.file_handler,
            df=df,
            transform=transform,
            target_transform=self.target_transform,
        )
```

autrainer/transforms/specific_transforms.py
```diff
@@ -300,12 +303,48 @@ def __init__(
...
         self.std = std
-        self._normalize = T.Normalize(mean=mean, std=std)
+        self._mean = torch.as_tensor(mean)
+        self._std = torch.as_tensor(std)
 
     def __call__(self, data: torch.Tensor) -> torch.Tensor:
-        if data.dtype == torch.uint8:
-            data = data.float()
-        return self._normalize(data)
+        views = {3: (-1, 1, 1), 2: (-1, 1), 1: (-1,)}
```
ATriantafyllopoulos, line 310: Why this check?

ATriantafyllopoulos:
> audobject records the arguments that are passed to a class during instantiation, meaning we cant serialize and deserialize the class later on when the concrete values depend on some internal computation and the dataset; basically we have to do it this way to trick audobject into thinking the class was instantiated as usual

Are you sure this is true? As far as I know, overwriting the values will work properly:

```python
>>> import audobject
>>> class Foo(audobject.Object):
...     def __init__(self, x: int):
...         self.x = x
... 
>>> bar = Foo(4)
>>> bar.x = 5
>>> audobject.from_yaml_s(bar.to_yaml_s())
{'$__main__.Foo': {'x': 5}}
```

ATriantafyllopoulos:
If I understand the current implementation correctly (can't really boil it down to a comment in a single line), I have the feeling that `Standardize` will now be computed separately for `train`, `dev`, `test`. The default should be to compute it once on the training set and assign it as part of the `dev`/`test` pipelines. Am I right or have I missed something?

autrainer/transforms/transform_manager.py
ATriantafyllopoulos, line 1: I really need some inline comments to follow along with what's going on in this file...(I also needed them in the past, just didn't get around to asking)

autrainer/transforms/specific_transforms.py
ramppdev: What do you mean in particular? Sorry i dont really get what this is referring to.

autrainer/transforms/smart_compose.py
ramppdev: Yeah this is also weird with the current setup. I need to think about a better solution.

autrainer/transforms/transform_manager.py
ramppdev: Sure, i will add some :)

ramppdev:
> > audobject records the arguments that are passed to a class during instantiation, meaning we cant serialize and deserialize the class later on when the concrete values depend on some internal computation and the dataset; basically we have to do it this way to trick audobject into thinking the class was instantiated as usual
> 
> Are you sure this is true? As far as I know, overwriting the values will work properly:
> 
> ```python
> >>> import audobject
> >>> class Foo(audobject.Object):
> ...     def __init__(self, x: int):
> ...         self.x = x
> ... 
> >>> bar = Foo(4)
> >>> bar.x = 5
> >>> audobject.from_yaml_s(bar.to_yaml_s())
> {'$__main__.Foo': {'x': 5}}
> ```

whoops :D i was trying to avoid doing exactly this all the time...
Well then my proposal would be to omit the `from_global` class method and transform and the global transform wrapper.
Then we could use a `setup` function that handles the correct calculation of e.g. the mean and std parameters.
However, the reason why i jumped right to the conclusion that audobject "replays" the initialization of the object was that they obviously need the initialization interface to reinstantiate the class. This would mean that we have to create a Standardize class that looks like the following:

```python
class StandardizeExample(audobject.Object):
    def __init__(self, x=None, y=None) -> None:
        self.x = x
        self.y = y
    
    def setup(self, ...) -> None:
        ... # calculate mean and std
        self.x = ...
        self.y = ...
```

The workflow would then be that the user specifies the transform in the transforms `autrainer.transforms.StandardizeExample` without passing any arguments. Then we call the setup function and calculate the relevant params. Then we can save the audobject and later when re reinstantiate it, we directly pass the already computed values.

This has the advantage that the weird switcheroo setup is not needed but is also kinda weird as we need these placeholder arguments.

> If I understand the current implementation correctly (can't really boil it down to a comment in a single line), I have the feeling that `Standardize` will now be computed separately for `train`, `dev`, `test`. The default should be to compute it once on the training set and assign it as part of the `dev`/`test` pipelines. Am I right or have I missed something?

Yes this is a big oversight, sorry. For some reason i completely missed the point that we need to add the same instance of the standardizer to all of the transform pipelines. This should be easily fixable by adding a `subset` param to the standardizer and passing the dataset itself to the setup function describes above. This would make sense anyways as it would align with our other setup function signatures: `def setup(self, data: "AbstractDataset") -> None:`.

The only issue is that i find it valuable (as described in the initial PR description) to have completely separate pipelines to be more flexible for modifications / deletions. This would mean that we need to compute the standardizer 3x if it would be specified for each subset.

@ATriantafyllopoulos what do you think about this direction? If you view it kinda similar i will open a PR to this one with the suggested changes so we can get a feeling for it.

autrainer/transforms/specific_transforms.py
ATriantafyllopoulos: Sorry, it was unclear to me what `views` does, can you please add an inline comment to state that the values will be kept (in `axes`) and used for an expansion?

autrainer/transforms/specific_transforms.py
```diff
@@ -300,12 +303,48 @@ def __init__(
         super().__init__(order=order)
         self.mean = mean
         self.std = std
-        self._normalize = T.Normalize(mean=mean, std=std)
```
ATriantafyllopoulos, line 303: Hmm, why not keep this? Are the two transforms interchangeable?

ramppdev: Hmm i was following the `Normalize` definition of [torchvision](https://pytorch.org/vision/main/generated/torchvision.transforms.Normalize.html). We actually have `ScaleRange` to scale between `[low, high]`. I would like to stick to this naming convention if possible.

We should probably define `Standardize` to be `Normalize` (torchvision) class in a global sense? Also see https://github.com/autrainer/autrainer/pull/82#issuecomment-2590480742.

> Hmm, why not keep this? Are the two transforms interchangeable?

I assume this refers to dropping `T.Normalize(mean=mean, std=std)`. The reason is that torchvision only supports images (e.g. 3D image and 4D batched image inputs), but we will encounter 1D-3D examples (see `views`) below. Hope this will be more clear once i add some inline comments.

ATriantafyllopoulos:
> Well then my proposal would be to omit the from_global class method and transform and the global transform wrapper.
Then we could use a setup function that handles the correct calculation of e.g. the mean and std parameters.
However, the reason why i jumped right to the conclusion that audobject "replays" the initialization of the object was that they obviously need the initialization interface to reinstantiate the class. This would mean that we have to create a Standardize class that looks like the following:
...

The workflow would then be that the user specifies the transform in the transforms autrainer.transforms.StandardizeExample without passing any arguments. Then we call the setup function and calculate the relevant params. Then we can save the audobject and later when re reinstantiate it, we directly pass the already computed values.

Yes, I think this would be the correct approach, though you would still need a way to manage the `setup`? Would you check if a transform has it to call it, or integrate it as a `no-op` in the base class and override it?

> This has the advantage that the weird switcheroo setup is not needed but is also kinda weird as we need these placeholder arguments.

Don't worry, these placeholder arguments are the canonical way to (ab)use `audobject` like this right from its origin

ATriantafyllopoulos:
> Yes this is a big oversight, sorry. For some reason i completely missed the point that we need to add the same instance of the standardizer to all of the transform pipelines. This should be easily fixable by adding a subset param to the standardizer and passing the dataset itself to the setup function describes above. This would make sense anyways as it would align with our other setup function signatures: def setup(self, data: "AbstractDataset") -> None:.

Yes, that the user specifies the subset on which the transform should be computed is fine, and will allow for a separate normalization of the different subset if the user wishes to do so (sometimes it is actually "needed").

> The only issue is that i find it valuable (as described in the initial PR description) to have completely separate pipelines to be more flexible for modifications / deletions. This would mean that we need to compute the standardizer 3x if it would be specified for each subset.

This is not ideal as it can take up to a few hours if the dataset is large (like `MSP-Podcast`). We can write some voodoo code to only run it on a subset if it has not already been run on it under-the-hood and still keep the same interface you want for the user. Do you want to do it like this?

ramppdev:
> Yes, I think this would be the correct approach, though you would still need a way to manage the `setup`? Would you check if a transform has it to call it, or integrate it as a `no-op` in the base class and override it?

Yup this would be the approach i had in mind, i would do it analogous to the optional criterions `setup(...)` or the `get_collate_fn(...)` which do not have to be specified in the base class and are just checked with a getattr call.

> This is not ideal as it can take up to a few hours if the dataset is large (like `MSP-Podcast`). We can write some voodoo code to only run it on a subset if it has not already been run on it under-the-hood and still keep the same interface you want for the user. Do you want to do it like this?

I mean yeah i would be open to it if we manage to write this half way decently but not sure how to do that without some weird borderline unmaintainable filters that would only be used for this kind of transform.

autrainer/transforms/specific_transforms.py
ramppdev: resolved; see below

autrainer/transforms/smart_compose.py
ramppdev: resolved; see below

ramppdev:
Alright, this is no longer draft, maybe a review from you @ATriantafyllopoulos would also be nice if you have time.
I reverted back most of the changes due and made the implementation way simpler:
- `base` part of the transforms is shared again, e.g. if you place a `Standardizer` in there, it will be calcuated only once
- using the `@Tag` syntax still allows you to override (or remove) specific tagged transforms if needed; but removing and replacing transforms only works in the respective subset (e.g., if you want to remove a dataset `train` transform, you have to explicitly do so in the model `train` transforms)
- transforms are setup in the dataset after it is initialized (`setup_transforms`), this has to be done after the dataset itself is fully initialized as some datasets do stuff after their init function (e.g. the toy dataset)
- setting up a global transform now has the same API as criterions, i.e., the dataset instance is passed to it and the transform can do whatever it wants with it; for `Standardizer`, this is currently implemented bare-bones, so the transform can only be used as the first one (as outlined in the docstring) but i think this is fine
- small (non-critical) bugs: incorrect variable naming in toy dataset wrapper, screwed up the upper bound of randint (now corrected to 256) which was used for tests and the toy dataset

autrainer/transforms/specific_transforms.py
```diff
@@ -321,12 +324,105 @@ def __init__(
...
+        subset: str = "train",
+        order: int = -101,
+    ) -> None:
+        """Standardize a dataset by calculating the mean and standard deviation
+        of the data and applying the normalization.
+
+        The mean and standard deviation are automatically calculated from the
+        data in the specified subset.
+
+        Note: The transform has to be applied as the first transform in the
```
ATriantafyllopoulos, line 363: This might be problematic for `serving.py` as it has to be coupled with the `preprocessing_transform`. Or do you think this is handled nicely with the present setup?

autrainer/transforms/transform_manager.py
```diff
@@ -53,139 +53,58 @@ def get_transforms(
...
-            model_transforms: List of model transforms.
-            dataset_transforms: List of dataset transforms.
-
-        Returns:
-            The filtered model and dataset transforms.
-        """
+        for transform in model_transforms:
+            key = self._get_key(transform)
+            seen.add(key)
+            if isinstance(transform, dict) and transform[key] is None:
```
ATriantafyllopoulos, line 82: Can you add an inline comment to describe what this if-statement does?

autrainer/transforms/transform_manager.py
```diff
@@ -53,139 +53,58 @@ def get_transforms(
...
-                if transform[key] is None and len(transform.keys()) == 1:
-                    l1.remove(transform)
-                    d_idx = self._find_matching_transform(l2, key)
-                    if d_idx is not None:
-                        l2.pop(d_idx)
-            return l1, l2
-
-        model_transforms, dataset_transforms = _filter(
-            model_transforms, dataset_transforms
+        combined.extend(
```
ATriantafyllopoulos, line 86: Note: is it clear from the docs that `model_transforms` will have priority over `dataset_transforms`?

autrainer/transforms/specific_transforms.py
```diff
@@ -321,12 +324,105 @@ def __init__(
...
+            )
+
+    def _init_transform(self) -> Optional[AbstractTransform]:
+        if self.mean and self.std:
+            return Normalize(self.mean, self.std, self.order)
+
+    def _collect_data(self, data: "AbstractDataset") -> torch.Tensor:
+        dataset_wrapper = getattr(data, f"{self.subset}_dataset")
+        original_transform = dataset_wrapper.transform
+        dataset_wrapper.transform = None
```
ATriantafyllopoulos, line 399: Why do we have to set it to `None`? For simplicity?

Cause I think that it might be useful to keep any previous transforms (e.g. one could presumably normalize features to range [0, 1] first and then make sure they have 0 mean and standard deviation of 1)

autrainer/transforms/transform_manager.py
ramppdev: Yes, we have it here: https://autrainer.github.io/autrainer/modules/transforms.html#online-transforms
> Model [online transforms](https://autrainer.github.io/autrainer/modules/transforms.html#online-transforms) outweigh dataset [online transforms](https://autrainer.github.io/autrainer/modules/transforms.html#online-transforms). If a model specifies a transform that is also specified in the dataset, the model transform is used, overriding the dataset transform.

autrainer/transforms/specific_transforms.py
ramppdev: Yes this would not be a problem as we keep the order of transforms the same by taking a two-step approach of first preprocessing and then applying the inference transform: https://github.com/autrainer/autrainer/blob/410f6aa2e203607b1db19f43d0af717b68adec22/autrainer/serving/serving.py#L336-L341

autrainer/transforms/transform_manager.py
ramppdev: See https://github.com/autrainer/autrainer/pull/82/commits/28d6abce58d77274f41953a7283666630a2be20f

autrainer/transforms/specific_transforms.py
ramppdev: Yeah this is here for simplicity right now:
> setting up a global transform now has the same API as criterions, i.e., the dataset instance is passed to it and the transform can do whatever it wants with it; for Standardizer, this is currently implemented bare-bones, so the transform can only be used as the first one (as outlined in the docstring) but i think this is fine

We can change it such that the pipeline is resolved up until the current standardizer instance? This would allow placing it anywhere.

ATriantafyllopoulos: > We can change it such that the pipeline is resolved up until the current standardizer instance? This would allow placing it anywhere.

No need for now, let's open an issue for improvement after merging, I think we can live with it as it is, so as not to delay this branch much longer

ramppdev: Sounds good, then i will try to get the docs ready ASAP.

ramppdev:
@ATriantafyllopoulos would be ready from my side :)

ATriantafyllopoulos (approved):

Issue #84, opened 2024-12-25
Per-Example Loss Values are Wrong for BalancedCrossEntropyLoss


ramppdev:
In our current implementation of the `BalancedCrossEntropyLoss` we calculate the class weights based on the inverse frequency of each class in the training subset, which is correct. However, we have an oversight that we normalize the weights to sum up to `1` instead of the number of classes `C`:
https://github.com/autrainer/autrainer/blob/c25132ee008508fb9cbd941724c6e9bfe86a0a4b/autrainer/criterions/criterion_wrappers.py#L28-L45

This is no issue when setting `reduction="mean"` which is the default for the [CrossEntropyLoss](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#torch.nn.CrossEntropyLoss) implementation, as torch automatically rescales the weights in this case (can also be seen in the formulas and [here](https://github.com/pytorch/pytorch/blob/2981534f54d49fa3a9755c9b0855e7929c2527f0/aten/src/ATen/native/LossNLL.cpp#L265-L267)).

**BUT** when obtaining the per-example loss again in our `OutputsTracker`, we end up with loss values that are essentially downscaled by a factor of `C`.
This is not "severe" in a sense that the loss computation is wrong during training, but all our reported per-example loss values are definitely not correct.

millinma:
I guess you are right. I don't seem to find that they specify this anywhere explicitly (if I didn't miss it), but I guess if we assume from the formulas that not specifying a weight results in a weight of 1, then everything should sum up to C. But I also agree that it shouldn't have impact on the training with reduction equal to "mean". For the curriculum scoring function, absolute values might be affected, but not the ranking right?

ramppdev:
> I guess you are right. I don't seem to find that they specify this anywhere explicitly (if I didn't miss it), but I guess if we assume from the formulas that not specifying a weight results in a weight of 1, then everything should sum up to C. But I also agree that it shouldn't have impact on the training with reduction equal to "mean".

I checked the C++ implementation and they explicitly normalize to the number of classes, but you can also easily verify this by passing different weights scaled by any constant other than 0 to the loss and compare the outputs when using `reduction="mean"`.
So this had absolutely no effect on the correctness of the training results and only affected the reporting of the per-example loss value npy files.


> For the curriculum scoring function, absolute values might be affected, but not the ranking right?

[...]

PR #86, opened 2024-12-29
Add More Balanced and Weighted Criterions

### LINKED ISSUE #42: Offer Weighted Criterions for all Training Tasks
autrainer currently only offers a weighted (class-balanced) criterion for multi-class classification at the moment.
This should be extended to offering balanced versions for regression and multi-label classification.
It could also be useful to add weighted versions with custom weights which would be passed in the dataset configuration using shorthand syntax.

ramppdev:
Closes #42 by adding:
- `BCEWithLogitsLoss` wrapper to handle float conversion and updating all multi-label configs
- `BalancedBCEWithLogitsLoss` automatically calculating class weights analogous to `BalancedCrossEntropyLoss` for multi-label classification
- `WeightedBCEWithLogitsLoss`, `WeightedCrossEntropyLoss`, and `WeightedMSELoss` (this one only for multi-target regression) to specify custom weights manually

There is also some debate on how to handle the weights calculation for multi-label classification using the `pos_weight` attribute of the [BCEWithLogitsLoss](https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html#torch.nn.BCEWithLogitsLoss) torch implementation. I chose to align the implementation with standard classification. Would you change it?

Also, for regression, there is no balanced variant as it is not really defined here but a manual reweighing of the multi-target outputs may make sense in some cases which is why i added the `WeightedMSELoss` implementation.

---

For example, for the `EmoDB` dataset, manual class weights could be specified like this:
```yaml
...
criterion:
  autrainer.criterions.WeightedCrossEntropyLoss:
    class_weights:
      anger: 1
      boredom: 2
      ...
...
```


autrainer/criterions/classification.py
```diff
@@ -0,0 +1,168 @@
...
+            data: Instance of the dataset.
+        """
+        frequency = (
+            data.df_train[data.target_column]
+            .map(data.target_transform)
+            .value_counts()
+            .sort_index()
+            .values
+        )
+        weight = torch.tensor(1 / frequency, dtype=torch.float32)
```
wagnerp4, line 44: 1 / frequency could potentially lead to division by zero for missing train samples

```suggestion
        frequency = (
            data.df_train[data.target_column]
            .map(data.target_transform)
            .value_counts()
            .sort_index()
            .values
        )
        if (frequency == 0).any():
    raise ValueError("Balanced weighting failed: at least one class/label has zero frequency.")
        weight = torch.tensor(1 / frequency, dtype=torch.float32)
```

wagnerp4 (commented):
# Observations:

CrossEntropy differs to MSE/BCE in weight registration:
- CrossEntropy Losses use **self.weight=weight** to store the class weights.
- BCE and MSE Losses use **self.register_buffer("weight", weight)** instead.
-> Potentially change CrossEntropy to use self.register_buffer("weight", weight)

CrossEntropy differs to MSE/BCE in shape/usage:
- CrossEntropy variants, the weighting is handled internally by PyTorch’s torch.nn.CrossEntropyLoss if self.weight is set.
- For BCE and MSE variants, the code multiplies the final output of the parent super().forward(x, y) by self.weight.expand_as(y).
-> For all Weighted/Balanced classes, consider adding a brief docstring note that “this class expects reduction='none' to be automatically set by autrainer so that weighting is applied on a per-example, per-label basis.”

# Comments:

Consider adding or ensure a test for zero-frequencies of Balanced Losses if relevant: I.e. in test_criterions.py create a small mock dataset with a label that never appears in df_train. For instance, dataset with labels [0, 1] but df_train only contains [0, 0, 0], e.g., label 1 has zero frequency.

**pos_weight**: Sticking to a frequency-based approach, mirroring balanced classification, keeps things consistent with CrossEntropyLoss. PyTorch’s built-in pos_weight could be used, but it only reweights the positive examples for each label, which doesn’t always match a “balanced” approach for datasets with many negative examples.

**WeightedMSELoss**: There is no well-defined “balanced MSE”; offering a manual weighting mechanism should be enough in most use cases. 

Maybe adding a short note in the docs about why chosing to skip a “BalancedMSELoss” and that PyTorch’s pos_weight behaves differently from the class reweighting we do in balanced classification can help users.

autrainer/criterions/classification.py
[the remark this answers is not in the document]
ramppdev: Yeah good point, i just assumed this would never be the case. This would apply to all criterions then.

ramppdev:
> CrossEntropy differs to MSE/BCE in weight registration:
> 
> * CrossEntropy Losses use **self.weight=weight** to store the class weights.
> * BCE and MSE Losses use **self.register_buffer("weight", weight)** instead.
>   -> Potentially change CrossEntropy to use self.register_buffer("weight", weight)

For the [cross entropy loss](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#torch.nn.CrossEntropyLoss) torch already provides a manual rescaling weight given to each class, this is not the case for BCE/MSE, which is why i use a buffer. This is whats happening under to hood for the CE loss as well.


> CrossEntropy differs to MSE/BCE in shape/usage:
> 
> * CrossEntropy variants, the weighting is handled internally by PyTorch’s torch.nn.CrossEntropyLoss if self.weight is set.
> * For BCE and MSE variants, the code multiplies the final output of the parent super().forward(x, y) by self.weight.expand_as(y).
>   -> For all Weighted/Balanced classes, consider adding a brief docstring note that “this class expects reduction='none' to be automatically set by autrainer so that weighting is applied on a per-example, per-label basis.”

Yeah sounds good to add it here as well, we automatically set `reduction="none"` and have this marked in the docs (not online currently as the docs are only updated for each release): https://github.com/autrainer/autrainer/blob/7a15ef7c6b74eef3d93750b742887c16ff243ed6/docs/source/modules/criterions.rst?plain=1#L13-L18

> **pos_weight**: Sticking to a frequency-based approach, mirroring balanced classification, keeps things consistent with CrossEntropyLoss. PyTorch’s built-in pos_weight could be used, but it only reweights the positive examples for each label, which doesn’t always match a “balanced” approach for datasets with many negative examples.
> 
> Maybe adding a short note in the docs about why chosing to skip a “BalancedMSELoss” and that PyTorch’s pos_weight behaves differently from the class reweighting we do in balanced classification can help users.

Yes this is a good idea :)

docs/source/modules/criterions.rst
```diff
@@ -30,20 +30,45 @@ For more information see `this discussion <https://discuss.pytorch.org/t/dataloa
...
 
-.. autoclass:: autrainer.criterions.BalancedCrossEntropyLoss
+.. autoclass:: autrainer.criterions.BCEWithLogitsLoss
    :members:
 
 .. autoclass:: autrainer.criterions.MSELoss
    :members:
 
+
+Balanced and Weigted Criterions
```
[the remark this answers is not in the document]
ramppdev: resolved in https://github.com/autrainer/autrainer/pull/86/commits/e83592a0883bbc32c1a1588bd2389a1d2df6e292

autrainer/criterions/classification.py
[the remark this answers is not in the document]
ramppdev: resolved in https://github.com/autrainer/autrainer/pull/86/commits/5f79174b89bc123d63f1cb9eb18150b9e059ac90

ramppdev:
I forgot to add the balanced MSE notes and the pos_weight text for BCE. Will do that tomorrow.

autrainer/criterions/classification.py
```diff
@@ -0,0 +1,177 @@
...
+        values = []
+
+        for label in data.target_transform.labels:
+            if label not in self.class_weights:
+                raise ValueError(f"Missing class weight for label '{label}'.")
+            values.append(self.class_weights[label])
+
+        assert_nonzero_frequency(np.array(values), len(data.target_transform))
+        weight = torch.tensor(values, dtype=torch.float32)
+        self.weight = weight * len(weight) / weight.sum()
```
ATriantafyllopoulos, line 84: Why `self.weight = weight/weight.mean()` and not `weight/weight.sum()`? Shouldn't it sum up to 1?

ATriantafyllopoulos (commented):
Do you think it's easy to add some correctness tests, i.e. manually compute some weight/loss values for some arbitrary targets/pseudopredictions?

autrainer/criterions/classification.py
```diff
@@ -0,0 +1,177 @@
...
+        weights.
+
+        Args:
+            x: Batched model outputs.
+            y: Targets.
+
+        Returns:
+            Loss.
+        """
+        return super().forward(x, y) * self.weight.expand_as(y)
```
ATriantafyllopoulos, line 142: Why not rely on `pos_weight` here? You could just compute the weights for each positive class vs its negative and then pass it to that attribute?

Or what is the difference in the two implementations?

autrainer/criterions/classification.py
```diff
@@ -0,0 +1,177 @@
...
+
+        for label in data.target_transform.labels:
+            if label not in self.class_weights:
+                raise ValueError(f"Missing class weight for label '{label}'.")
+            values.append(self.class_weights[label])
+
+        assert_nonzero_frequency(np.array(values), len(data.target_transform))
+        weight = torch.tensor(values, dtype=torch.float32)
+        weight = weight * len(weight) / weight.sum()
+        self.register_buffer("weight", weight)
```
ATriantafyllopoulos, line 177: Similar to the above, perhaps we can rely on `pos_weight`?

autrainer/criterions/classification.py
ramppdev: No, actually not, for reference see https://github.com/autrainer/autrainer/issues/84 and https://github.com/autrainer/autrainer/pull/85.

When using cross entropy, passing a `weight` tensor, and using `reduction=`"mean"`, passing a tensor that sums to 1, the number of classes, or anything else does not matter as torch automatically scales the weights accordingly (see ref).

BUT: when you pass `reduction=`"none"`, it is important that the weights sum up to the number of classes (see on torch website). You can easily check this by manually doing `.mean()` and see that the calculation is off by the class scaling factor.

ramppdev:
> Do you think it's easy to add some correctness tests, i.e. manually compute some weight/loss values for some arbitrary targets/pseudopredictions?

I mean we can compute normal loss values without reduction and then manually rescale them and compare the output? Not sure if this would provide a lot more assurance that the code is correct?

autrainer/criterions/classification.py
ramppdev: see above.

autrainer/criterions/classification.py
ramppdev: The idea to not use pos_weight is that `pos_weight` is only available for the BCE losses, and not for MSE / CE obviously, i.e. for MSE we need this trick anyways, and full consistency between the way the balancing and weighting is done for all losses.

autrainer/criterions/classification.py
ATriantafyllopoulos: Can you please add this as inline comment? And also link to the place in the torch docs where this is stated because I can't find it in v2.5.0:https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html

autrainer/criterions/classification.py
ATriantafyllopoulos: But what happens if the user attempts to provide `pos_weight` because this is what they are familiar with from native `torch`? It will be propagated to the constructor and `work`, won't it? So then there are two ways to do the same thing?

ATriantafyllopoulos:
> I mean we can compute normal loss values without reduction and then manually rescale them and compare the output? Not sure if this would provide a lot more assurance that the code is correct?

No, I meant compute them with pen and paper ;)

So:

```python
target: [0, 1]
output: [0.3, 0.7]
weight: [1, 1]
---> loss: .5146
###########
target: [0, 1]
output: [0.3, 0.7]
weight: [1, 2]
---> loss: 1.0292
```

and so on

autrainer/criterions/classification.py
ramppdev: This is actually never stated in the docs, i found it in the cpp implementation: https://github.com/pytorch/pytorch/blob/2981534f54d49fa3a9755c9b0855e7929c2527f0/aten/src/ATen/native/LossNLL.cpp#L265-L267

autrainer/criterions/classification.py
ramppdev: Yeah good point. So should i either figure out how to use `pos_weight` while still achieving essentially the exact same balancing scheme or disable it my internally overriding it to none?

autrainer/criterions/classification.py
ramppdev: Will resolve this as the decision from above goes for all.

ramppdev:
> > I mean we can compute normal loss values without reduction and then manually rescale them and compare the output? Not sure if this would provide a lot more assurance that the code is correct?
> 
> No, I meant compute them with pen and paper ;)
> 
> So:
> 
> ```python
> target: [0, 1]
> output: [0.3, 0.7]
> weight: [1, 1]
> ---> loss: .5146
> ###########
> target: [0, 1]
> output: [0.3, 0.7]
> weight: [1, 2]
> ---> loss: 1.0292
> ```
> 
> and so on

Haha unfortunately my budget is a bit tight so i wont be able to acquire any pens or papers until next month, maybe @wagnerp4 can help out and manually compute these :D (if not no problem i will do it on the weekend then or sth)

autrainer/criterions/classification.py
ATriantafyllopoulos: What exactly do you want to achieve that `pos_weight` doesn't do by default?

ATriantafyllopoulos:
> Haha unfortunately my budget is a bit tight so i wont be able to acquire any pens or papers until next month, maybe @wagnerp4 can help out and manually compute these :D (if not no problem i will do it on the weekend then or sth)

Hehe, let's skip it for now and add an issue for @wagnerp4 to work on after this has been merged

autrainer/criterions/classification.py
ramppdev: The thing is, `pos_weight` does only scale the positive (see https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html). This makes things cumbersome and inconsistent with how we handle class imbalance for other losses (like CE or MSE), which do not have a pos_weight option. Matching the same factor on both terms would require extra logic (a per-class global weighting plus a carefully chosen `pos_weight`) i.e. it will be very weird code to match the loss of the current version with the loss produced when using `pos_weight`.
I would either completely refactor the approach such that there is no "balanced" version of BCE or use `#negative / #positive` for the weight as torch suggests (but then only for the BCE variants and keep CE / MSE the same).

ramppdev: @ATriantafyllopoulos can this be merged now or do you want to use `pos_weight` over the current implementation?

ATriantafyllopoulos: Yes, go ahead and merge

Issue #92, opened 2025-01-13
Undefined features_path


millinma:
Shouldn't `pop` return `None` here by default, as otherwise a `KeyError` is raised if `features_path` is not defined (which I think should be optional)?

https://github.com/autrainer/autrainer/blob/7a15ef7c6b74eef3d93750b742887c16ff243ed6/autrainer/core/scripts/preprocess_script.py#L64

ramppdev:
Oh yes good point this is wrong.

PR #94, opened 2025-01-14
Fix Undefined Features Path in Preprocessing Script

millinma:
Quick fix for #92 adding default return `None` when reading features_path in preprocessing.


ramppdev:
Thanks looks great :)

ramppdev (approved):

PR #96, opened 2025-01-15
Fix Device Fallback if not Compiled with Cuda

ramppdev:
Small fix that properly handles when torch is not compiled with cuda. If the requested device is not available, a `RuntimeError` is thrown. However, if torch is not compiled with cuda at all, an `AssertionError` is thrown.


wagnerp4 (commented):
Includes `AssertionError` in device selection to explicitely handle the Torch not compiled with CUDA error

PR #97, opened 2025-01-18
Remove Checkpoint Attributes from Model Config during Fetching

### LINKED ISSUE #93: model_checkpoint in fetch script
There seems to be an issue in the instantiation if `model_checkpoint` is defined in the model config, as it this is passed to the instantiation, potentially causing issues (especially with the torchvision wrapper)

https://github.com/autrainer/autrainer/blob/7a15ef7c6b74eef3d93750b742887c16ff243ed6/autrainer/core/scripts/fetch_script.py#L76

so I would suggest to add before:
```
model.pop("model_checkpoint", None)
```

ramppdev:
Closes #93 by removing the checkpoint-related attributes from the model configuration which should not be passed to the model as parameters during instantiation (see [model docs](https://autrainer.github.io/autrainer/modules/models.html)).

### DIFF
--- a/autrainer/core/scripts/fetch_script.py
+++ b/autrainer/core/scripts/fetch_script.py
@@ -73,6 +73,10 @@ class FetchScript(AbstractPreprocessScript):
         for name, model in self.models.items():
             print(f" - {name}")
             model.pop("transform", None)
+            model.pop("model_checkpoint", None)
+            model.pop("optimizer_checkpoint", None)
+            model.pop("scheduler_checkpoint", None)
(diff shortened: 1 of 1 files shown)


millinma (approved):
LGTM

wagnerp4:
Looks good! This change should prevent inherited default args or non-implemented config parameters to interfere with the transfer logic

wagnerp4 (commented):
Looks good! This change should prevent inherited default args or non-implemented config parameters (besides transfer) to run in the fetch script

PR #98, opened 2025-01-18
Expand Quickstart Docs with List and Show Examples

### LINKED ISSUE #81: Info on changing the data path
Maybe it would make sense to point here towards the (IMO common) use case of changing the data path of a dataset. 
https://github.com/autrainer/autrainer/blob/c069542243ce1ae5817b9d0b6eb096bbcaa000cb/docs/source/usage/quickstart.rst?plain=1#L20

If I am not mistaken, the intended way is to do `autrainer create --empty` and then `autrainer show <dataset_id> -s`, followed by changing the `conf/dataset/<dataset_id>.yaml`. I think this one in particular should be a super common use case  very frequently and to me doesn't become clear quickly, without diving deeper into the docs.

ramppdev:
Closes #81 by outlining the workflow from config creation, over discovery, to overriding local configurations (e.g., with a custom dataset path) in the autrainer quickstart.

Also fixes the following minor bugs:
- `e.g.,` was written without a comma in lots of places
- the lexer for autrainer commands did not handle dashes or equals very well, leading to a change in highlighting

### DIFF
--- a/docs/extensions/sphinx_lexers.py
+++ b/docs/extensions/sphinx_lexers.py
@@ -1,7 +1,15 @@
 from typing import Any, Dict
 
 from pygments.lexer import RegexLexer, bygroups
-from pygments.token import Keyword, Name, Number, Operator, String, Text
+from pygments.token import (
+    Keyword,
+    Literal,
+    Name,
+    Number,
+    Operator,
+    String,
+    Text,
+)
 from sphinx.application import Sphinx
 from sphinx.highlighting import lexers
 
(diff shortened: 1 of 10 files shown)


docs/extensions/sphinx_lexers.py
wagnerp4, line 1: Added equals to accomodate key-value pairs and visually hightlight them with `pygments.token.Literal` usage.
Expanded parsing to dash-preceding text:  **(r"[^-\s]+", Text)** -> **(r"[^\s]+", Text)**

docs/source/usage/quickstart.rst
wagnerp4, line 1: Typo in l.97: To perform a grid search over multiple ~multiple~ configurations [...]

ramppdev: Good catch :) Fixed in https://github.com/autrainer/autrainer/pull/98/commits/d1900230ea9777c0f93635e413fc881a182106a1

ramppdev:
@millinma could you check again if this is enough for you?

PR #100, opened 2025-01-21
Fix/dataset dataclass

ATriantafyllopoulos:
This provides some fixes for #61:

1. Renaming the main argument in all `AbstractModel` classes to match what the `create_model_input` func expects
2. Change `bookkeeping` to use `create_model_input` as the `save_model_summary` wouldn't work if there were more inputs
3. Added a `default_collate_fn` parameter to `AbstractDataset` to more easily allow the user to override it and specify their custom `collate_fn`

In general, I like your solution with the `signature` checking. I think it makes things more easily extensible while remaining mostly under-the-hood (it does require the user to specify `features` as the name of their model input, I think we should write a neater error message if that does not happen)

Once we merge this I can rebase #61 to `main` and put the finishing touches

### DIFF
--- a/autrainer/core/utils/bookkeeping.py
+++ b/autrainer/core/utils/bookkeeping.py
@@ -15,7 +15,7 @@ from autrainer.metrics import AbstractMetric
 from autrainer.models.utils import create_model_inputs
 
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
     from autrainer.datasets.utils import DataItem, DatasetWrapper
 
 
--- a/autrainer/core/utils/bookkeeping.py
+++ b/autrainer/core/utils/bookkeeping.py
@@ -12,11 +12,11 @@ from torchinfo import summary
 import yaml
 
 from autrainer.metrics import AbstractMetric
-from autrainer.training.utils import create_model_inputs
+from autrainer.models.utils import create_model_inputs
 
 
 if TYPE_CHECKING:
-    from autrainer.datasets import AbstractDataset  # pragma: no cover
+    from autrainer.datasets.utils import DataItem, DatasetWrapper
 
 
 class Bookkeeping:
@@ -105,22 +105,20 @@ class Bookkeeping:
     def save_model_summary(
         self,
         model: torch.nn.Module,
-        dataset: "AbstractDataset",
+        dataset: Union[List["DataItem"], "DatasetWrapper"],
         filename: str,
     ) -> None:
         """Save a model summary to a file.
 
         Args:
             model: Model to summarize.
-            dataset: Dataset to get the input size from.
+            dataset: Data to get the input sizes from.
+                Either a batch or a list of data points
             filename: Name of the file to save the summary to.
         """
-        batch = next(iter(dataset.train_loader))
-        model_inputs = create_model_inputs(model, batch)
+        model_inputs = create_model_inputs(model, dataset[0])
         x = [
-            np.expand_dims(
-                getattr(dataset.train_dataset[0], key), axis=0
-            ).shape
+            np.expand_dims(getattr(dataset[0], key), axis=0).shape
             for key in model_inputs.keys()
         ]
 
--- a/autrainer/core/utils/bookkeeping.py
+++ b/autrainer/core/utils/bookkeeping.py
@@ -12,6 +12,7 @@ from torchinfo import summary
 import yaml
 
 from autrainer.metrics import AbstractMetric
+from autrainer.training.utils import create_model_inputs
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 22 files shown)


autrainer/core/utils/bookkeeping.py
```diff
@@ -12,10 +12,11 @@
 import yaml
 
 from autrainer.metrics import AbstractMetric
+from autrainer.models.utils import create_model_inputs
 
 
 if TYPE_CHECKING:
```
ramppdev, line 18: claim your free + 0.00001% coverage :)
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

autrainer/core/utils/bookkeeping.py
```diff
@@ -104,17 +105,22 @@ def create_folder(self, folder_name: str, path: str = "") -> None:
     def save_model_summary(
         self,
         model: torch.nn.Module,
-        dataset: "AbstractDataset",
+        dataset: Union[List["DataItem"], "DatasetWrapper"],
```
ramppdev, line 108: When i did the trainer init "speedup", i refactored this function a good bit:
https://github.com/autrainer/autrainer/blob/7ca06a52c784d1c5b223e2918c9a325b6534885b/autrainer/core/utils/bookkeeping.py#L100-L136

Just passing in the `shape` would simplify the implementation a lot here, so could be adapted here as well :)

autrainer/datasets/abstract_dataset.py
```diff
@@ -276,7 +280,7 @@ def train_loader(self) -> DataLoader:
             shuffle=True,
             generator=self._generator,
             collate_fn=self.train_transform.get_collate_fn(
-                self, default=DataBatch.collate
+                self, default=self.default_collate_fn
```
ramppdev, line 283: since we pass a `self` reference to `get_collate_fn` anyways, we could remove explicitly passing `default=self.default_collate_fn`.
The only reason why this would be interesting is if we want to pass different collation functions for different subsets.

But if we really need this, wouldnt it be cleaner to do have sth like this as the signature:
```python
def get_collate_fn(self, data: "AbstractDataset", default_collate: Optional[Callable] = None) -> Callable:
   default_collate = default_collate or data.default_collate_fn
```
Meaning that we can simplify the interface of the train loader. One less thing the user has to explicitly pass and potentially mess up.

autrainer/models/utils.py
```diff
@@ -185,3 +188,15 @@ def embeddings(self, x: torch.Tensor) -> torch.Tensor:
         self.model(x)
         self._unregister()
         return self._embeddings
+
+
+def create_model_inputs(
+    model: AbstractModel,
+    data: Union[AbstractDataBatch, DataItem],
+) -> Dict[str, torch.Tensor]:
+    _forbidden = ["features", "target", "index"]
```
ramppdev, line 197: Should we make this public? Or will these 3 always stay the same as "illegal" inputs?

Also i had no check in there that features is actually the first one. I think this has to be enforced.

ramppdev (commented):
I would say we are very close to finishing this :)

autrainer/core/utils/bookkeeping.py
ramppdev: Yeah also possible :)

autrainer/datasets/abstract_dataset.py
ATriantafyllopoulos: Indeed, I think it makes sense to remove the `default` argument, I did so in 7190324

autrainer/models/utils.py
ATriantafyllopoulos: > Should we make this public? Or will these 3 always stay the same as "illegal" inputs?

I think we should stay strict with these three for now as they will otherwise break a lot of our logic in `Trainer`.

> Also i had no check in there that features is actually the first one. I think this has to be enforced.

I added a check in `AbstractModel.inputs` 639b9d4

ATriantafyllopoulos:
> I would say we are very close to finishing this :)

I think we are now good to merge (to the base branch, where I'll review again and prepare for merging to `main`)

autrainer/models/abstract_model.py
```diff
@@ -35,4 +35,15 @@ def inputs(self) -> List[str]:
         Returns:
             Model inputs.
         """
-        return [v.name for v in signature(self.forward).parameters.values()]
+        names = [v.name for v in signature(self.forward).parameters.values()]
+        if names[0] != "features":
+            raise NameError(
```
ramppdev, line 40: There are some typos here:
```python
(
    f"Model {type(self).__name__} "
    "does not have 'features' "
    "as the first argument of its 'forward' method. "
    f"Its arguments are: {names}. "
    "Please rewrite the 'forward' method accordingly."
)
```
Also interesting that ruff doesnt complain about the unnecessary f-strings

ramppdev (approved):
Looks good :)

Issue #88, opened 2025-01-09
Integration of sound event detection


ATriantafyllopoulos:
This is a master issue to track the integration of sound event detection.

+ [x] Start a new branch (feature/sed). We will develop on this branch and make new PR for each of the bullet points that follow. https://github.com/autrainer/autrainer/tree/feature/sed
+ [ ] Integrate data interface and sample dataset [DESED](https://zenodo.org/records/3702397)
+ [ ] Integrate basic model, the baseline from DCASE will do for a start
+ [ ] Integrate loss
+ [ ] Integrate metrics
+ [ ] Add tests

PR #101, opened 2025-01-31
Issue #88: Integrate data interface and sample dataset DESED

wagnerp4:
Integrate data interface and sample dataset DESED:

Introduces following new classes:
- `SegmentedDatasetWrapper`: extension to `DatasetWrapper`, implementing a static download function that is currently being called in the `DCASE2019Task4.download` function
- `DCASE2019Task4` with `DCASE2019Task4-32k.yaml` for spectrograms after preprocessing: This class implements static download function that currently is missing the real recording dataset, which is officially being created by calling `youtube-dl` package. How should we address this difference in the downloading process (compared to zenodo)?
- `BaseSEDDataset` for SED problems

Following new changes:
- `ZipDownloadManager._extract_archive` extended for downloading necessary .tar.gz files from zenodo

Current problems:
1. Using e.g. CNN10 with segment_wise_path we have 3-dimensional output causing issues with tracker, that expects 2-dimensional output
2. Windowing currently happens on download with static call to `SegmentedDatasetWrapper` needs to improved (still missing: bbox annotations, currently only offers windowing on download)


ramppdev:
I think the base branch is outdated, the PR removes some stuff that is probably unwanted.

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -90,3 +90,236 @@ def __getitem__(
...
+            target = torch.tensor(float(target), dtype=torch.float32)
+        if self.transform is not None:
+            data = self.transform(data, index=item).float()
+        if self.target_transform is not None:
+            target = self.target_transform(target)
+
+        return data, target, item
+
+    @staticmethod
+    def create_fixed_windows(
```
ramppdev, line 172: What is the reasoning to have this function in the dataset wrapper instead of the dataset itself? seems like it is only called in the download method of the dataset so wouldnt it make sense to place it as a private function there?

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -90,3 +90,236 @@ def __getitem__(
...
+                )[0]
+                data = torch.from_numpy(data)
+            except Exception as e:
+                print(f"Crashed with {e} at {item_path}")
+                exit()
+        else:
+            data = self.file_handler.load(item_path)
+
+        target = self.df.loc[index, self.target_column]
+        if isinstance(target, str) and target.startswith("[["):
```
ramppdev, line 160: Should a target starting with  `[[...` map to a list of lists that we need to map back to a tensor? I would like to avoid calling eval here.

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -90,3 +90,236 @@ def __getitem__(
...
+                    item_path,
+                    offset=self.df.loc[index, "start"],
+                    duration=self.df.loc[index, "end"]
+                    - self.df.loc[index, "start"],
+                    always_2d=True,
+                )[0]
+                data = torch.from_numpy(data)
+            except Exception as e:
+                print(f"Crashed with {e} at {item_path}")
+                exit()
```
ramppdev, line 155: Can be removed if we intend to fail anyways.

autrainer/datasets/utils/dataset_wrapper.py
wagnerp4: I agree, as of right now the windowing is just related to downloading, and is best put there. The reason was, that every sed dataset class would need the same (private) windowing function, this seemed counter-intuitive to me for code separation. Also I thought maybe users want to apply a re-window maybe as part of a pre-processing transform - but it a single 'perfect' windowing on download seems good too.

autrainer/datasets/abstract_dataset.py
```diff
@@ -658,7 +599,137 @@ def __init__(
...
+            min_event_gap: Minimum gap between events in seconds
+            threshold: Threshold for classification
+        """
+        self._assert_threshold(threshold)
+        self.threshold = threshold
+        super().__init__(
+            path=path,
+            features_subdir=features_subdir,
+            seed=seed,
+            task="ml-classification",
```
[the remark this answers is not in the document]
wagnerp4: Yes true, I changed it

autrainer-configurations/dataset/DCASE2019Task4-32k.yaml
```diff
@@ -0,0 +1,18 @@
...
+seed: ${seed}
+index_column: filename
+target_column: segment_events
+file_type: npy
+file_handler: autrainer.datasets.utils.NumpyFileHandler
+
+criterion: torch.nn.BCEWithLogitsLoss
+metrics:
+  - autrainer.metrics.multi_label_classification.MLF1Weighted
+tracking_metric: autrainer.metrics.multi_label_classification.MLF1Weighted
```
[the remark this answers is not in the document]
wagnerp4: Changed. This will change in the future, after integrating the sed-related metrics

autrainer/datasets/abstract_dataset.py
```diff
@@ -105,18 +93,9 @@ def __init__(
         self.stratify = stratify or []
 
         self._generator = torch.Generator().manual_seed(self.seed)
+        self.df_train, self.df_dev, self.df_test = self.load_dataframes()
         self._assert_stratify()
 
-    @property
```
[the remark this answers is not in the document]
wagnerp4: I fetched and rebased to main, this fixed it, but there are still 3 errors which I will look into

autrainer/datasets/utils/dataset_wrapper.py
ramppdev: Then i would say we have 2 options:
1. Since all SED datasets inherit from `BaseSEDDataset`, the base class can handle the windowing which is then accessible in all SED datasets.
2. If windowing should (also) be accessible as a preprocessing step, it should be implemented as a transform. This comes a the cost of managing (and not forgetting) to apply the preprocessing transform everywhere, but at the same time provides more functionality and freedom.

The first approach would seem more logical to me. BUT I dont know a lot about SED, so you should take this with a big grain of salt.

autrainer-configurations/dataset/DCASE2019Task4-32k.yaml
```diff
@@ -0,0 +1,20 @@
...
+criterion: torch.nn.BCEWithLogitsLoss
+metrics:
+  - autrainer.metrics.MLAccuracy
+  - autrainer.metrics.MLF1Micro
+  - autrainer.metrics.MLF1Macro
+  - autrainer.metrics.MLF1Weighted
+tracking_metric: autrainer.metrics.MLF1Weighted
+
+transform:
+  type: grayscaley
```
ramppdev, line 20: `grayscaley` (y at the end) is one reason for the tests failing.

autrainer/datasets/abstract_dataset.py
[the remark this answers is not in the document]
ramppdev: Ah the `feature/sed` branch is also not up to date with master :D You probably have to rebase it as well now.

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -57,8 +58,7 @@ def __init__(
 
     def _create_file_path(self, file: str) -> str:
         path = Path(self.path, self.features_subdir, file)
-        if self.file_type is not None:
-            path = path.with_suffix(f".{self.file_type}")
+        path = path.with_suffix(f".{self.file_type}")
```
ramppdev, line 61: This is the reason the 2 tests are failing, we had this to be nullable as we want to make as few assumptions as possible on the actual format during preprocessing:
https://github.com/autrainer/autrainer/blob/21b8ffaadba5a347e07d7370440bdc6a85bae5ea/autrainer/core/scripts/preprocess_script.py#L59

During "normal" training, this is always required and set to the `file_type` passed to the dataset.

autrainer/datasets/utils/dataset_wrapper.py
wagnerp4: I agree. I suggest using the BaseSEDDataset for now. This also makes sense because we can also overwrite _assert_target_column there to deal with different targets. I have changed the code to it.

autrainer/datasets/utils/dataset_wrapper.py
wagnerp4: Removed "[[" and "eval" code. We now have a flattened segment_event vector e0_s0, ..., e39_s9 (multi-label setting).

autrainer/datasets/utils/dataset_wrapper.py
wagnerp4: Changed, thank you

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -90,3 +91,74 @@ def __getitem__(
...
+            window_size: Size of the segmented window in seconds. Ignored if data is pre-windowed.
+            hop_size: Hop size in seconds. Ignored if data is pre-windowed.
+            min_event_length: Minimum duration of an event in seconds. Ignored if data is pre-windowed.
+            min_event_gap: Minimum duration between events in seconds. Ignored if data is pre-windowed.
+        """
+        super().__init__(**kwargs)
+        self.is_prewindowed = (
+            "start" in self.df.columns and "end" in self.df.columns
+        )
+        if not self.is_prewindowed:
```
ATriantafyllopoulos, line 118: I don't see how these attributes are used below..?

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -90,3 +91,74 @@ def __getitem__(
...
+        If data is already windowed (has 'start' and 'end' columns), windowing parameters are ignored.
+
+        Args:
+            window_size: Size of the segmented window in seconds. Ignored if data is pre-windowed.
+            hop_size: Hop size in seconds. Ignored if data is pre-windowed.
+            min_event_length: Minimum duration of an event in seconds. Ignored if data is pre-windowed.
+            min_event_gap: Minimum duration between events in seconds. Ignored if data is pre-windowed.
+        """
+        super().__init__(**kwargs)
+        self.is_prewindowed = (
```
ATriantafyllopoulos, line 115: I would not hardcode this but rather make the names of the columns be arguments that can be manipulated via `hydra`

autrainer/datasets/utils/dataset_wrapper.py
```diff
@@ -90,3 +91,74 @@ def __getitem__(
...
+            if window_size is None:
+                raise ValueError(
+                    "window_size must be provided for non-windowed data"
+                )
+            self.window_size = window_size
+            self.hop_size = hop_size
+            self.min_event_length = min_event_length
+            self.min_event_gap = min_event_gap
+
+    def __getitem__(
```
ATriantafyllopoulos, line 128: As the only thing that's changing is how we load the audio file, I would abstract this as a new method in the base class, and only override this part here, to avoid duplicating the code for the labels

heoj1N: I agree, this should be in `BaseSEDDataset`

autrainer/datasets/utils/dataset_wrapper.py
wagnerp4: Right, I adapted it to the other classes, and changed the dataset configuration file to include the columns start and end

autrainer/datasets/utils/dataset_wrapper.py
wagnerp4: I agree, this should be in BaseSEDDataset

ATriantafyllopoulos (commented):
I stopped the review midway and decided to re-implement it myself as there were too many changes I wanted to ask. Have a look at my comments and hold your breath for my own PR

ATriantafyllopoulos:
Closing in favor of #119

PR #110, opened 2025-02-27
Fix Callback Positioning for Loggers and Docs

ramppdev:
This is a collection of multiple small bugfixes related to our loggers:
- the file saving-related threads need to be joined before `cb_on_train_begin`, as loggers (and other stuff) can access these (this is the first callback)
- same goes with saving the `metrics.csv`, this has to be done before `cb_on_train_end`
- the continue training callback has to be last in the callback order such that it can replay potential loggers etc.
- during aggregation, we replay the newly created run which potentially modifies the params passed to it, so we need to copy them properly
- added mlflow and tensorboard loggers to tests to catch this behavior in the future
- added full class paths to the logger docs such that they are properly resolved to internal links
- install all extras in the CD pipeline when building the docs such that optional documentation is properly rendered and not the fallback behavior, see e.g., mlflow logger docs: https://autrainer.github.io/autrainer/modules/loggers.html#autrainer.loggers.MLFlowLogger

### DIFF
--- a/.github/workflows/CD.yaml
+++ b/.github/workflows/CD.yaml
@@ -32,7 +32,7 @@ jobs:
       - name: Install Poetry
         uses: snok/install-poetry@v1
       - name: Install Dependencies
-        run: poetry install --with docs
+        run: poetry install --all-extras
       - name: Build Docs
         run: poetry run sphinx-build -b html docs/source docs/build
       - name: Publish to GitHub Pages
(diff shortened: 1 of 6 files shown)


ATriantafyllopoulos (approved):
It's not clear to me why any of that was happening but I trust you :)

Maybe add some failure examples before the fixes for reference

ramppdev:
> It's not clear to me why any of that was happening but I trust you :)
> 
> 
> 
> Maybe add some failure examples before the fixes for reference

All of these things were happening because of skill issues on my side :D i will check what i can cover now with tests but most things cant happen anymore so i would need to mock behavior that is not possible anymore.

PR #111, opened 2025-02-28
Add Optional DataLoader Arguments

### LINKED ISSUE #16: Optional DataLoader Arguments
In some scenarios it is useful to configure additional dataloader arguments such as the number of workers or pinning memory.
Therefore allow for the specification of `train_loader_kargs` (same for dev and test).

The question is if this should be specified in the main configuration file or in each dataset configuration?
I would suggest the main configuration file, as e.g. the number of workers is also tied to the hardware.

ramppdev:
Closes #16 by changing the loader properties to methods with explicit arguments that can be overridden.
These are specified as `${subset}_loader_kwargs` overrides that can either be placed globally in the `conf/config.yaml` or in the dataset configuration for each subset. The dataset configuration takes precedence over the global one, allowing for full control.

This also moves the (inference) batch size specification to the loader methods such that the dataset init is debloated.

TODO:
- [x] update docs with new `${subset}_loader_kwargs` overrides


ATriantafyllopoulos (approved):
All good, please make it clear in the docs that these can be specified, and how :)

ramppdev:
Wasnt all good :D the following changes were also required:
- probability generators for augmentations need to be different across workers (but still based off the dataset seed) to ensure that they are not doing the same
- since we now have functions creating the dataloaders, we can drop `setup_transforms` for simplicity
- handle pinned memory transfers correctly
- pop loader kwargs during dataset preprocessing if present
- tests

ATriantafyllopoulos (approved):

PR #117, opened 2025-03-17
feat: Integrate basic SED models (CRNN, FDY-CRNN, HTSAT)

wagnerp4:
Integrating basic SED models as part of #88


ramppdev:
I am a bit confused, is this already included in #118?

PR #118, opened 2025-03-17
Feature/sed metric

wagnerp4:
## Description
This PR integrates the core metrics for Sound Event Detection (SED) evaluation as part of the SED feature implementation (#88).

### Changes
- Added Event-based F1 score
- Added Segment-based F1 score
- Added Segment-based Error Rate (SER) metric
- Added metrics tests

### TODO (Future PRs)
- [ ] Add Median Filtering implementation
- [ ] Add PSDS1 score from DCASE

### Related
- Closes part of #88 (SED feature)
- SER metric reference: https://arxiv.org/pdf/1710.02997v1


autrainer-configurations/model/Crnn.yaml
ramppdev, line 1: The naming should be `CRNN.yaml` to be consistent with the class name, same goes for the other model config :)

ramppdev (commented):
Will do a proper review once all TODOs that should be in here are resolved and the tests pass :)

PR #119, opened 2025-03-18
WIP: Feature/sed updated

ATriantafyllopoulos:
As promised in #101 I opened a new PR for #88.

Counter to my own advice, I opened a behemoth PR that includes changes in

a) The dataset, beginning from #101 
b) The model, taken from #117
c) And the metrics (not taken from #118)

I have also rebased on the latest `main` to benefit from recent changes.

This is a WIP with numerous open issues. However, I have hopefully simplified things by:

1. Relying on [SED_EVAL](https://tut-arg.github.io/sed_eval/tutorial.html#sed-eval-evaluation-toolbox-for-sound-event-detection) for evaluation. While not perfect, this is the codebase most other researchers rely on. We can improve our interface to it by calling some base classes if needed
2. I now no longer refactor the dataset at `download` but instead do so during code. I think this has vastly simplified the implementation. I am not yet finished, but we now avoid this confusing manipulation of the dataframe. Instead, we create a new dataframe at runtime which has a single column per file and a `np.array` with the ground truth segments (conditioned on `frame_rate` which needs to be provided)
3. I have adapted `OutputsTracker` to handle sequential targets/predictions

@wagnerp4 please review this PR **thoroughly**. Note down the changes to your code and give me feedback. I think this can form a much better foundation for improving this branch and eventually merging straight to `main` once we are ready.


autrainer/datasets/dcase_2019_t4.py
```diff
@@ -0,0 +1,206 @@
...
+            file_type: File type of the features.
+            file_handler: File handler to load the data.
+            batch_size: Batch size.
+            inference_batch_size: Inference batch size. If None, defaults to
+                batch_size. Defaults to None.
+        """
+        self.onset_column = "onset"
+        self.offset_column = "offset"
+        # FIXME: make more modular
+        self.duration = 10  # duration fixed @ 10s
```
wagnerp4, line 83:
```suggestion
        self.duration = DURATIONS["length_sec"]
```

autrainer/datasets/dcase_2019_t4.py
```diff
@@ -0,0 +1,206 @@
...
+        self.frame_rate = frame_rate
+        assert self.frame_rate is not None
+        assert self.frame_rate > 0
+        # FIXME: Maybe np.ceil?
+        self.num_frames = int(self.duration / self.frame_rate)
```
wagnerp4, line 88:
```suggestion
        self.num_frames = int(np.ceil(self.duration / self.frame_rate))
```

wagnerp4 (commented):
This branch makes things easier. I took a look at the changes and left comments and questions. I also think this state is a much better starting point, simplifying label handling to synthetic data, metric usage to `sed_eval`, and the use of `SEDEncoder` for windowing/annotation handling. Combined with the benefit of current main state.

Todos:
- The sequential nature of the data and intermediate shapes of models and model configuration for sed cause some problems during training. This could imply multiple higher-level code change impact, such as with the change of the `OutputsTracker`, for the future.
- The baseline `CRNN` model's implementation still has some weaknesses, which have to be adressed.
- Improved parameterization and usage of class arguments

ATriantafyllopoulos:
@wagnerp4 I rebased on the latest main after merging #61

Please apply your fixes to the above issues as a PR on this branch starting by the current point. We will do a final rebasing before merging, but all new changes we push to the main now should not impact things much

autrainer/metrics/sed_metrics.py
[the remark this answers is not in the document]
ATriantafyllopoulos: Yea, exactly, I want to find a way to avoid rerunning the same code multiple times to get another metric from `sed_eval`. Can you think of a way to do it with a static object?

autrainer/models/crnn.py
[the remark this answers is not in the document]
ATriantafyllopoulos: So `CNN10` is not geared towards `SED` because of its multiple `pooling` layers. That's fine, we can skip it for now.

For `CRNN`, the key is to abstract away its instantiation so that the user can specify the downsampling rate and match it to the frame rate of the dataset. Currently, there's a 3-way dependency:

a) The frame rate in which the dataset returns its labels. This is set in the dataset config

b) The hop size of the features, set in the preprocessing config or transform

c) The model pooling sizes (in the time dimension), set in the model config

b) and c) need to be fixed so that they jointly match a). This is not clear right now. I think it is okay to demand this from the user in the first version. Which means:

I) We should document it properly

II) We should test for this and fail gracefully if it is violated (i.e. with an  informative error)

autrainer/datasets/dcase_2019_t4.py
ATriantafyllopoulos: It is a question...Should we use `np.ceil` or `np.round`?

wagnerp4: np.ceil ensures we never cut off audio, but it might include an extra frame.
I would say it's better to use np.ceil here to fill up potential events at the end

autrainer/metrics/sed_metrics.py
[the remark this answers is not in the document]
ramppdev: Since we always evaluate the whole list of metrics one after another the following would work:
- use a singleton "backend" and implement the sed metrics once there, then store them in a cache (dict)
- create a "frontend" metric class that follows our normal interface but delegates the computation to the backend and just receives its key if already computed
- introduce a `clear()` function that is called once after evaluating all metrics and invalidates the cache

This is not very clean, but would expose compatible frontend classes while reducing double computation if the metric calculation is expensive.

ATriantafyllopoulos:
Closing as stale, might revisit SED in the future

PR #122, opened 2025-03-26
Add AudioSet Dataset

### LINKED ISSUE #121: AudioSet integration

ATriantafyllopoulos:
Closes #121 

Now finished the integration and tested locally with a few training runs.

There are a couple of open points that I'd like to discuss:

1) Tests -- I can't write any meaningful ones as all the functionality depends on having the `CSV` files and `ontology.json` available. I would rather not add another dependency on external assets (that need to be downloaded for the test). On the other hand, it is quite important to test `include/exclude`. So maybe I'll do so (please review while I think about it and let me know what you think)

2) There is some "undesired" behavior regarding the selection of the final labels as I've implemented it. [PANNs](https://github.com/qiuqiangkong/audioset_tagging_cnn/blob/master/metadata/class_labels_indices.csv) used a custom blacklist which selects a subset of the labels. I am not sure why, but ever since, several papers are referencing their 527 labels. The way I do it, labels are determined based on availability (i.e. after filtering for available files and applying include/exclude filters). I end up with 557 if I don't use include/exclude. I am wondering whether I should also offer the option for the 527, though I can't think of a nice way to include it in the dataset, other than by adding the list of labels as part of the code or creating a `yaml` version with these classes specified as `target_column`


ramppdev:
@ATriantafyllopoulos we can create drafts PRs now since the repo is public :)

ATriantafyllopoulos:
> pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file.

@ramppdev I get this locally too. Any idea why? I just added a dependency in `pyproject.toml`

ramppdev:
> > pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file.
> 
> 
> 
> @ramppdev I get this locally too. Any idea why? I just added a dependency in `pyproject.toml`

Yes this is completely normal. Poetry tracks the deps in the `pyproject.toml` by pinning them exactly in the `poetry.lock` file. If you change the dependencies you have to also adapt the lockfile accordingly by running `poetry lock`.

autrainer/datasets/audioset.py
```diff
@@ -0,0 +1,470 @@
...
+            path: Root path to the dataset.
+            features_subdir: Subdirectory containing the features.
+                If `None`, defaults to audio subdirectory,
+                which is `default` for the standard format,
+                but can be overridden in the dataset specification.
+            seed: Seed for reproducibility.
+            metrics: List of metrics to calculate.
+            tracking_metric: Metric to track.
+            index_column: Index column of the dataframe.
+            target_column: Target column of the dataframe.
```
ramppdev, line 119: We should note here that the target column can be `None`.

autrainer-configurations/dataset/AudioSet-wav.yaml
```diff
@@ -0,0 +1,25 @@
+id: AudioSet
+_target_: autrainer.datasets.AudioSet
+
+threshold: 0.5
+
+path: data/AudioSet2021/
+features_subdir: ""
```
ramppdev, line 7: From below it seems like the `features_subdir` should always be `""` / `None`?

autrainer/datasets/audioset.py
```diff
@@ -0,0 +1,470 @@
...
+            df = pd.concat((df, u_df))
+        return df
+
+    @cached_property
+    def df_dev(self) -> pd.DataFrame:
+        return self._load_df("eval_segments")
+
+    @cached_property
+    def df_test(self) -> pd.DataFrame:
+        return self.df_dev
```
ramppdev, line 453: is dev == test for audioset?

ATriantafyllopoulos: It is for me :)

ATriantafyllopoulos:
> Maybe there is some minimal mocking possible (i.e. similar to the other dataset tests) but i agree that it is not super nice to include lots of assets only for testing.

I don't really like "fake" mock tests. I wanted to try, but it was very hard for me to create the CSVs in that weird format that audioset comes with.

> I guess the easiest would be to just link the csv file for this option and require the user to manually download it and place it alongside the dataset files? This way, we could check if class_labels_indices.csv is present?

I'll leave that for a future update of the dataset so as to move the release forward.

autrainer/datasets/audioset.py
ATriantafyllopoulos: There's no other available development set

autrainer-configurations/dataset/AudioSet-wav.yaml
ATriantafyllopoulos: No, it might be set to some other value in case the user wants to preprocess the data

ramppdev: Ah yes you are absolutely right

ramppdev:
> > Maybe there is some minimal mocking possible (i.e. similar to the other dataset tests) but i agree that it is not super nice to include lots of assets only for testing.
> 
> I don't really like "fake" mock tests. I wanted to try, but it was very hard for me to create the CSVs in that weird format that audioset comes with.
> 
> > I guess the easiest would be to just link the csv file for this option and require the user to manually download it and place it alongside the dataset files? This way, we could check if class_labels_indices.csv is present?
> 
> I'll leave that for a future update of the dataset so as to move the release forward.

I agree, then lets create an issue for it and move forward :)

PR #123, opened 2025-03-27
Added CCCLoss

ATriantafyllopoulos:
This PR adds the `CCCloss`. It is a similar implementation to the one used to train https://huggingface.co/audeering/wav2vec2-large-robust-12-ft-emotion-msp-dim but adapted to match the biased std/var estimations of `numpy` so it's compatible with `audmetric`

### DIFF
--- a/autrainer/criterions/regression.py
+++ b/autrainer/criterions/regression.py
@@ -2,6 +2,7 @@ from typing import TYPE_CHECKING, Dict
 
 import numpy as np
 import torch
+from torch.nn.modules.loss import _Loss
 
 from .utils import assert_nonzero_frequency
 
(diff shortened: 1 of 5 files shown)


autrainer/criterions/regression.py
```diff
@@ -85,3 +86,70 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+        corr = torch.sum(v_pred * v_gt, axis=0) / (
+            torch.sqrt(torch.sum(v_pred**2, axis=0))
+            * torch.sqrt(torch.sum(v_gt**2, axis=0))
+        )
+        sd_gt = torch.std(y, axis=0, unbiased=False)
+        sd_pred = torch.std(x, axis=0, unbiased=False)
+        numerator = 2 * corr * sd_gt * sd_pred
+        denominator = var_gt + var_pred + (mean_gt - mean_pred) ** 2
+        ccc = numerator / denominator
+        ccc = ccc.mean()
```
ramppdev, line 148: In order to record the per-example losses, each loss needs to return the non-reduced loss vector of shape `(b,)`, see: 
https://github.com/autrainer/autrainer/blob/8d86aa969d30209ed0c25c11c06d42791bb975d1/autrainer/training/training.py#L127-L131

tests/test_criterions.py
```diff
@@ -295,6 +299,78 @@ def test_forward_dtype(self) -> None:
...
+                    [0.3, 0.3, 0.3],
+                ),
+                torch.Tensor(
+                    [0.3, 0.3, 0.3],
+                ),
+                marks=pytest.mark.xfail(raises=AssertionError),
+            ),
+        ],
+    )
+    def test_forward_dtype(self, x, y) -> None:
```
ramppdev, line 347: more or less checks the correctness rather than the dtype? Also typing :D

ATriantafyllopoulos: True, this is blind copy-pasting

autrainer/criterions/regression.py
ATriantafyllopoulos: Unfortunately this is not possible for `CCCLoss` as its computation is defined over the batch dimension.

I.e. if you have single-target regression, with a target vector Y and an output vector X, the `CCCLoss` is 1 minus the concordance correlation between X and Y.

It is simply not defined for each element.

---

Hmm. It seems like this is not compatible with `autrainer` in its present form and would need bigger changes. For calling the `.mean()` before `.backward()` we might be safe since `torch.tensor(1.).mean()` will work and give the same output (`torch.tensor(1)`) and the outputs tracker will store losses that do not match the length of the training. Therefore, I guess we won't get any errors, but it feels a bit wrong to abuse this flexibility. Do you have any views on this?

ramppdev: Yeah calling `mean` repeatedly is fine, but when tracking the "per-example" losses, these will run out of sync and only have the length of the batches instead of examples.

The workaround solution would be to just do `loss.expand(y.size(0))` and get a view (not a copy) of the tensor and then call mean for the backward pass and track batch-level losses for each example.
Kinda wrong but also kinda right if CCC is only defined over n examples?

PR #124, opened 2025-03-27
Add support for audb datasets

ATriantafyllopoulos:
This PR adds support for audEERING's [audb](https://audeering.github.io/audb/) and [datasets](https://audeering.github.io/datasets/datasets.html)

It is implemented as follows:
1. The user specifies the dataset `name` and the parameters they want for `audb.load`
2. The user specifies a `train_table`, `dev_table`, and `test_table`. These contain should contain the targets
3. The user specifies a `target_column` as usual. It can be a `string` or a `list`. The `task` and `target_transform` are computed on the fly based on the format of these columns. They must be present in all tables, otherwise the code will crash.

Open points:
1. The automatic `download` in `autrainer` is not compatible with the data management of `audb`, as our [fetch script](https://github.com/autrainer/autrainer/blob/8d86aa969d30209ed0c25c11c06d42791bb975d1/autrainer/core/scripts/fetch_script.py#L66) does a static instantiation of the dataset, which does not allow us to pass the necessary arguments to `audb.load`. I think this is fine for now.
2. Having a look at the [available datasets](https://audeering.github.io/datasets/datasets.html), very few of them match our requirements of having at least a set of very clear train/dev/test tables to draw from. This means that the benefit of supporting `audb` is maybe not worth the trouble of integrating this and maintaining it. I would see no big problem if the API is stable (will ask the maintainers)

### DIFF
--- a/autrainer/datasets/audb_wrapper.py
+++ b/autrainer/datasets/audb_wrapper.py
@@ -1,0 +1,274 @@
+from functools import cached_property
+from typing import Dict, List, Optional, Union
+
+import audb
+import audformat
+from omegaconf import DictConfig
+import pandas as pd
+
+from autrainer.datasets.abstract_dataset import AbstractDataset
+from autrainer.datasets.utils import (
+    AbstractTargetTransform,
+    LabelEncoder,
+    MinMaxScaler,
+    MultiLabelEncoder,
+    MultiTargetMinMaxScaler,
+)
+from autrainer.transforms import SmartCompose
+
+
+class AudbDataset(AbstractDataset):
+    def __init__(
+        self,
+        path: str,
+        features_subdir: str,
+        train_table: str,
+        dev_table: str,
+        test_table: str,
+        name: str,
+        audb_params: Dict,
+        seed: int,
+        metrics: List[Union[str, DictConfig, Dict]],
+        tracking_metric: Union[str, DictConfig, Dict],
+        target_column: str,
+        file_type: str,
+        file_handler: Union[str, DictConfig, Dict],
+        features_path: Optional[str] = None,
+        train_transform: Optional[SmartCompose] = None,
+        dev_transform: Optional[SmartCompose] = None,
+        test_transform: Optional[SmartCompose] = None,
+        stratify: Optional[List[str]] = None,
+        threshold: Optional[float] = None,
+    ) -> None:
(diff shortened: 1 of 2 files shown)


autrainer/datasets/audb_wrapper.py
```diff
@@ -0,0 +1,274 @@
...
+                minimum=self.db.tables[self.train_table]
+                .df[self.target_column]
+                .min(),
+                maximum=self.db.tables[self.train_table]
+                .df[self.target_column]
+                .max(),
+            )
+
+    @staticmethod
+    def download(path: str) -> None:  # pragma: no cover
```
ramppdev, line 206: I mean this kinda defeats the purpose of downloading the dataset beforehand to avoid race conditions when using parallel launchers during training. We could also change the download methods to take in the dataset `cfg` which would allow the direct download.

Here we call the download method: https://github.com/autrainer/autrainer/blob/8d86aa969d30209ed0c25c11c06d42791bb975d1/autrainer/core/scripts/fetch_script.py#L64-L69

We could do:
```python
    def _download_datasets(self) -> None:
        print("Fetching datasets...")
        for name, dataset in self.datasets.items():
            print(f" - {name}")
            if not dataset.get("path", None):
                continue
            hydra.utils.instantiate(
                config={
                    "_target_": dataset["_target_"] + ".download",
                    "_recursive_": False, # otherwise we try to pass the dataset instance
                    "cfg": dataset,
                },
            )  # pragma: no cover
```
and 
```python
    @staticmethod
    def download(cfg: DictConfig) -> None:  # pragma: no cover
        cfg.path # "new" way to get the path
```

ATriantafyllopoulos:
```python
    @staticmethod
    def download(cfg: DictConfig) -> None:  # pragma: no cover
        cfg.path # "new" way to get the path
```

It doesn't feel so nice to me to change the entire interface for `download` just so that we support `audb`, which is basically not used by anyone.

I mean, it is potentially useful to have access to all of the attributes in `cfg`, but perhaps it will be confusing to the inexperienced user?

ATriantafyllopoulos:
Closing as stale, `audb` hasn't really increased in popularity so I don't see the point to integrate this right now.

We can revisit if there's interest

PR #128, opened 2025-03-31
Updated SED dataset, metrics, and model

wagnerp4:
This PR references: https://github.com/autrainer/autrainer/pull/119#issuecomment-2750964074 and introduces some fixes to the prior implementation

- `BaseSEDMetricMixin` handles event-wise or segment-wise errors using `sed_eval` and array-based caching mechanism, see: https://github.com/autrainer/autrainer/pull/119#discussion_r2021196298.
- `PadCollate` transform that allows users to use the baseline architecture without changing their preprocessing/audio parameters. (this can potentially be removed and as disussed here: https://github.com/autrainer/autrainer/pull/119#discussion_r2011907791)
- Adapted baseline crnn.py to work. Included Batch-Norm layer that improves learning. Removed additional sigmoid from computation.
- `dcase2019_t4.py` now includes full `DURATIONS` dict with the dataset-specific duration parameters, that is passed to the `SEDEncoder`.
- Added small test suite for SED metrics


autrainer/metrics/sed_metrics.py
```diff
@@ -1,32 +1,246 @@
+import hashlib
+import threading
+from typing import Dict, List
+
 import numpy as np
-from sed_eval.sound_event import EventBasedMetrics
+from sed_eval.sound_event import EventBasedMetrics, SegmentBasedMetrics
+
+from ..datasets.utils.target_transforms.sed_encoder import SEDEncoder
```
ATriantafyllopoulos, line 8: Best import this from package, i.e. `from autrainer.datasets.utils.target_transforms import SEDEncoder`

autrainer/metrics/sed_metrics.py
ATriantafyllopoulos, line 1: I don't understand what's the purpose of this cache you introduced. It's part of the `Base` class. So it will be instantiated anew for each metric object you are creating. Will it save any computation?

autrainer/models/crnn.py
```diff
@@ -218,204 +212,31 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return self.cnn(x)
 
 
-# FDYCRNN model
-
-
-class FDYCRNN(AbstractModel):
+class BiGRU(nn.Module):
```
ATriantafyllopoulos, line 215: I don't think you really need this class as you simply pass the args to `nn.GRU`

autrainer/models/crnn.py
```diff
@@ -218,204 +212,31 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-
-
-# Shared
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        recurrent, _ = self.rnn(x)
+        return recurrent
 
 
-class BiGRU(nn.Module):
+class BiLSTM(nn.Module):
```
ATriantafyllopoulos, line 234: Same here, I don't think the class is needed

autrainer/transforms/specific_transforms.py
```diff
@@ -722,3 +720,41 @@ def __init__(
     def __call__(self, data: torch.Tensor) -> torch.Tensor:
         data = self.smile.process_signal(data.numpy(), self.sample_rate)
         return torch.from_numpy(self.smile.to_numpy(data)).squeeze()
+
+
+class PadCollate(AbstractTransform):
```
ATriantafyllopoulos, line 725: How is that different from [RandomCrop](https://audtorch.readthedocs.io/en/0.2.0/api-transforms.html#randomcrop) which is already available?

ATriantafyllopoulos: Also, the problem with `RandomCrop` is that you also have to adapt the `labels`. How do you handle that here?

ATriantafyllopoulos:
We decided to close this PR in favor of smaller ones with individual fixes

PR #130, opened 2025-04-01
v0.6.0

ATriantafyllopoulos:
@ramppdev it's been a while since I've done a release in GitHub. Should I bump the package version via this and then make a new release via the GitHub Web Interface, or what's the recommended process?


ramppdev:
Yes exactly, the version in the `pyproject.toml` has to be updated as well. After that, creating a new release (+ `v0.6.0` tag) triggers the pypi and docs deployment.

PR #132, opened 2025-04-05
Fix Mutable Hydra Multirun Configurations

### LINKED ISSUE #113: New job "overwrites" previous job’s configuration paths with --config-path
I noticed an issue when specifying a configuration folder using the following example:
`uv run autrainer train --config-path /path/to/conf`
where the `conf` folder follows the expected structure (as shown in the image)

Image.

The issue occurs when submitting a second job with a different path -- specifically, after updating the `results_dir` in `config.yaml` and the `path` in the dataset YAML file. Instead of each job maintaining its designated paths, the results of the first job are seemingly overwritten and stored in the new path of the second job. This also becomes clear when looking at the specified `results_dir` of the first job, which is then empty:

Image

What I expected:
- Each job should respect the `results_dir` and `path` values specified in its respective configuration **at submission time**.
- Previously submitted jobs should not be affected by newly submitted ones.

What I observed:
- The `results_dir` of the first job is different after submitting the second job.
- The first job does not store the results at its specified path anymore, but only has empty folder structures.

I don’t recall experiencing this issue before specifying `--config-path` explicitly in the CLI. Previously, I was able to change paths without affecting already running jobs.
I haven’t yet had the chance to look deeper into the issue, but I wanted to report this in case others are encountering the same problem.

For now, I avoid this issue by creating separate config folders for each experiment. This ensures that each job has its own configuration and does not interfere with previously started jobs.
`uv run autrainer train --config-path /path/to/conf_1`
`uv run autrainer train --config-path /path/to/conf_2`

ramppdev:
Closes #113 by copying all Hydra related configurations to a temporary directory before launching the runs.
Instead of exposing the `conf` directory (or the one specified with `--config-dir/-cd`), the temporary directory is added to the Hydra directories.
This way, edits to the currently executed configuration no longer have any effect on the actual multirun.

At the same time this also fixes:
- reformatting of main configurations to no longer include placeholder (`???`) paramters; this allows running useful hydra commands that need to compose the config like `autrainer train --info`
- handle incompatible `--config-path/-cp` Hydra argument and fail if this is present
- proper handling of setting matplotlib to `Agg` backend (this got somehow triggered again; i.e., i didnt fix it properly earlier)
- move `DataBatch` to local scope otherwise all CLI commands are slower

### DIFF
--- a/autrainer/core/scripts/utils.py
+++ b/autrainer/core/scripts/utils.py
@@ -144,7 +144,8 @@ def check_invalid_config_path_arg(parser: argparse.ArgumentParser) -> None:
             "The `--config-path/-cp` Hydra CLI argument is not supported "
             " by autrainer.\nTo override the config directory and name, "
             "use the `--config-dir/-cd` and `--config-name/-cn` arguments "
-            "instead.\nFor more information, refer to the Hydra documentation: "
-            "https://hydra.cc/docs/advanced/hydra-command-line-flags/"
+            "instead.\nFor more information, refer to the Hydra documentation "
+            "(https://hydra.cc/docs/advanced/hydra-command-line-flags/) and "
+            "https://github.com/autrainer/autrainer/pull/132."
         ),
     )
--- a/autrainer/core/scripts/utils.py
+++ b/autrainer/core/scripts/utils.py
@@ -1,3 +1,4 @@
+import argparse
 from functools import wraps
 import sys
 from typing import Any, Callable, Dict, Optional, TypeVar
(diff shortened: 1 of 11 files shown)


autrainer/core/scripts/utils.py
```diff
@@ -128,3 +129,22 @@ def add_hydra_args_to_sys(
...
+    if args.config_path is None:
+        return
+    raise CommandLineError(
+        parser=parser,
+        message=(
+            "The `--config-path/-cp` Hydra CLI argument is not supported "
+            " by autrainer.\nTo override the config directory and name, "
+            "use the `--config-dir/-cd` and `--config-name/-cn` arguments "
+            "instead.\nFor more information, refer to the Hydra documentation: "
+            "https://hydra.cc/docs/advanced/hydra-command-line-flags/"
```
ATriantafyllopoulos, line 148: Perhaps link to #132 so they know why we decided on this

ATriantafyllopoulos (commented):
In general I am okay with this, however I am not sure if it will interact with our `sweeper` plugin or with `submitit`.  Are we sure those don't do some weird copy of the `conf` dir themselves? I'm completely ignorant, just asking

autrainer/core/scripts/utils.py
ramppdev: See https://github.com/autrainer/autrainer/pull/132/commits/ca19fb10d5eadc9444cf72b741b63705dd3725c0

ramppdev:
> In general I am okay with this, however I am not sure if it will interact with our `sweeper` plugin or with `submitit`. Are we sure those don't do some weird copy of the `conf` dir themselves? I'm completely ignorant, just asking

This is totally fine and all sweepers and launchers are compatible. We redirect the config directory the same way `--config-dir/-cd` does, which happens even before the hydra main function is even called. So conceptually, sweepers and launcher do not even know of this change. The cleanup of the directory is also done after all jobs have been finished in the main process, so even when using slurm this is handled properly.
Ref to the [Hydra docs](https://hydra.cc/docs/experimental/callbacks/) for `on_multirun_end`: "When using a launcher, this will be executed on local machine."

PR #133, opened 2025-04-06
Add Transfer Argument to Models

### LINKED ISSUE #29: Serving pipeline downloads original model state in transfer learning
Currently, we used `audobject` to store the model during training and create it again during serving. However, this downloads again the original state if the model used transfer learning. This is unnecessary for the user who just wants to use the model.

I am not sure how uniform the code is, if all classes use the same argument (I doubt it), then we can override it when loading the `yaml` during serving.

P.S. We probably need a `needs improvement` label for this.

ramppdev:
Closes #29 by adding a `transfer` argument to all models. This allows for handling the transfer argument as follows:
- `autrainer fetch` and `autrainer train`: remove `transfer` if a model checkpoint is provided -> no weights need to be downloaded
- `autrainer inference`: override `transfer` to `None` as we will always load a checkpoint

Testability to verify this functionality is very limited and still kinda up to debate. Of course i checked everything manually but i dont have a good automated solution to do so right now.

### Additional changes
- This adds the `transfer` functionality to all our models, allowing to load all of them with pretrained weights. The only exceptions are `AudioRNNModel`, `FFNN`, and `SeqFFNN`. These will raise an error when you pass a truthy value for `transfer`. The reason they also have this argument now it to have a uniform interface for `AbstractModel`.

- Since we now have the ability to load all models with `transfer=None`, most tests will do so, mostly resolving the issue that tests may fail due to downloading model states here: https://github.com/autrainer/autrainer/pull/46#issuecomment-2752115788
For `TorchvisionModel`, we still keep the transfer test as `efficientnet_b0` is a small model state and verifies that overriding the last layer works.

### Breaking changes
Whisper, TDNN, and W2V2 models can now be initialized with random weights, so all of their pretrained-configs (ending with `-T`) were either newly created or include `transfer: true` now.

### DIFF
--- a/autrainer-configurations/model/ASTModel.yaml
+++ b/autrainer-configurations/model/ASTModel.yaml
@@ -1,0 +1,9 @@
+id: ASTModel
+_target_: autrainer.models.ASTModel
+
+transform:
+  type: raw
+  base:
+    - autrainer.transforms.FeatureExtractor:
+        fe_type: AST
+        fe_transfer: MIT/ast-finetuned-audioset-10-10-0.4593
--- a/autrainer-configurations/model/TDNNFFNN.yaml
+++ b/autrainer-configurations/model/TDNNFFNN.yaml
@@ -1,7 +1,6 @@
-id: TDNNFFNN-T
+id: TDNNFFNN
 _target_: autrainer.models.TDNNFFNN
 hidden_size: 32
-transfer: true
 
 transform:
   type: raw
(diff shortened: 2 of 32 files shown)


ATriantafyllopoulos (approved):
Thank you, this breaking change was something I wanted to add anyway, so that's totally fine. Can you please add 2 more `yaml` files for training `w2v2-b.yaml` and `w2v2-l.yaml`  from scratch?

ramppdev:
> Thank you, this breaking change was something I wanted to add anyway, so that's totally fine. Can you please add 2 more `yaml` files for training `w2v2-b.yaml` and `w2v2-l.yaml` from scratch?

These already exist, so i guess thats fine?

PR #137, opened 2025-04-08
Add metrics code

wagnerp4:
Applied fixes to dataset class, referencing https://github.com/autrainer/autrainer/pull/119.

This implementation uses multi-class inheritance of sed metrics from BaseAscending/DescendingMetric and SEDMetric. SEDMetric offers the implementation for event-based and segment-based metrics from sed_eval package. Metric classes are defined independently on module-level like in the other metric classes.

The implementation is missing a smart avoidance of re-computations for the individual metrics (._sed, ._sed.reset) for different metrics of the same type of either "event", "segment".

### DIFF
--- a/autrainer/metrics/sed_metrics.py
+++ b/autrainer/metrics/sed_metrics.py
@@ -1,32 +1,162 @@
 import numpy as np
-from sed_eval.sound_event import EventBasedMetrics
+from sed_eval.sound_event import EventBasedMetrics, SegmentBasedMetrics
 
-from .abstract_metric import BaseAscendingMetric
+from autrainer.datasets.utils.target_transforms import SEDEncoder
+
+from .abstract_metric import BaseAscendingMetric, BaseDescendingMetric
 
 
-class SegmentBasedF1(BaseAscendingMetric):
-    def __init__(self, target_transform):
-        """F1 macro metric using `sklearn.metrics.f1_score`."""
-        super().__init__(
+class SEDMetric:
+    def __init__(
+        self,
+        target_transform: SEDEncoder,
+        t_collar: float,
+        percentage_of_length: float,
+        time_resolution: float,
+        metric_type: str = "event",
+    ):
+        """SED Metric Mixin Class for Event-based and Segment-based metrics.
+        Uses `sed_eval.sound_event.{EventBasedMetrics, SegmentBasedMetrics}`.
+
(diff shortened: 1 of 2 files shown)


autrainer/metrics/sed_metrics.py
```diff
@@ -72,91 +102,85 @@ def _evaluate_events(self, y_true: np.ndarray, y_pred: np.ndarray) -> None:
...
+        backend = SEDMetricBackend()
+        backend.configure(
+            target_transform=target_transform, metric_type="segment", **kwargs
         )
-        BaseAscendingMetric.__init__(
-            self,
+        super().__init__(
             name="segment-based-f1",
-            fn=self.forward,
+            fn=lambda y_true, y_pred: backend(y_true, y_pred, "f_measure"),
```
ramppdev, line 158: instead of defining a lambda function, you can also directly pass `metric="f_measure"` to the super init call, all metrics pass through any additional keywords to the fn call.

pyproject.toml
```diff
@@ -55,6 +55,7 @@ psutil = "^5.0.0"
 PyYAML = "^6.0.1"
 Requests = "^2.31.0"
 scikit-learn = "^1.3.1"
+sed_eval = { version = "^0.2.1", optional = true }
```
ramppdev, line 58: If you add new dependencies to the `pyproject.toml`, please run `poetry lock` to generate the lockfile.

ramppdev: this is also the reason why the CI pytest check failed

ramppdev:
@wagnerp4 ah nice were you able to resolve the dependency issue?

wagnerp4:
> @wagnerp4 ah nice were you able to resolve the dependency issue?

I think so yes, the tests went through aswell! I saw some weird new dependencies in the diff to the old .lock file though. I'm not sure if that's usual.

ramppdev:
> > @wagnerp4 ah nice were you able to resolve the dependency issue?
> 
> 
> 
> I think so yes, the tests went through aswell. I saw some weird new dependencies in the diff to the old .lock file though. I'm not sure if that's usual. I will push again to remove the lamdba function.

Yes that is normal and expected as most of our deps are pinned tona range rather than an exact version for flexibility. There is a way to keep them mostly fixed with --no-update or sth like that but im usually also lazy and just run poetry lock.

PR #138, opened 2025-04-08
SED: Update CRNN model

wagnerp4:
Applied fixes to dataset class, referencing #119.

Removes module-level dependencies in favor of torch-integrated classes (for bidirectional `GRU`/`LSTM`). 

The model currently uses default arguments as in the baseline implementation. This causes dimensionality problems when not using a custom collate function, unless the user specifies the CRNN architecture accordingly using hydra, as discussed in: https://github.com/autrainer/autrainer/pull/119#discussion_r2011907791.

Removes reference to FDYCRNN model

### DIFF
--- a/autrainer/models/crnn.py
+++ b/autrainer/models/crnn.py
@@ -23,6 +23,7 @@ class CRNN(AbstractModel):
         hidden_size: int = 64,
         n_layers_rnn: int = 1,
         rnn_type: str = "LSTM",
+        bidirectional: bool = True,
     ) -> None:
         """DCASE 2019 Task 4 Baseline CRNN model. For more information see:
         Paper: https://inria.hal.science/hal-02160855.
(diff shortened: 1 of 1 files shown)


autrainer/models/crnn.py
```diff
@@ -99,56 +93,61 @@ def __init__(
...
 
-        def clip_grad_hook(grad):
-            return torch.clamp(grad, -self.fc_grad_clip, self.fc_grad_clip)
+        if rnn_type.upper() == "GRU":
+            self.rnn = nn.GRU(
+                input_size=cnn_output_channels,
+                hidden_size=hidden_size,
+                num_layers=n_layers_rnn,
+                batch_first=True,
+                bidirectional=True,
```
ATriantafyllopoulos, line 103: This should be a param (also below)

PR #141, opened 2025-04-16
Automatically Increment Generator Seed in Augmentation Pipeline

ramppdev:
This adds support for automatically incrementing the generator seed across a list of augmentations specified in a `AugmentationPipeline`. Lets say some pipeline is specified with multiple augmentations that have the same probability:
```yaml
...
generator_seed: 0
pipeline:
  - autrainer.augmentations.Some:
      p: 0.5
  - autrainer.augmentations.Random:
      p: 0.5
  - autrainer.augmentations.Aug:
      p: 0.5
```
Previously, this would apply either all augmentations or none with a 50/50 chance as the global generator seed for all augmentations would be set to 0. I think it makes more sense to interpret the global generator seed more as a starting point for the augmentation pipeline and increment the seed for each one that does not have an explicit generator seed.

**Note:** This is a breaking change, but i believe this change is more what users expect to happen in the first place.

### DIFF
--- a/autrainer/augmentations/augmentation_pipeline.py
+++ b/autrainer/augmentations/augmentation_pipeline.py
@@ -11,6 +11,7 @@ class AugmentationPipeline:
         self,
         pipeline: List[Union[str, Dict[str, Any]]],
         generator_seed: int = 0,
+        increment: bool = True,
     ) -> None:
         """Initialize an augmentation pipeline.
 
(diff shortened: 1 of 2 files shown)


ATriantafyllopoulos (approved):
Thanks! Yes, indeed, that's what I was personally assuming was happening

PR #147, opened 2025-05-02
Migrate Transforms and Augmentations to Operate on Structs 

ramppdev:
This migrates both transforms and augmentations to structs, such that their `__call__` methods take in `AbstractDataItem` instances.
Both transforms and augmentations now have the ability to access and modify the labels, or even implement custom logic based on the type of data being processed (i.e., do some additional stuff for SED when the time dimension is altered).

Whats not included here is a way to handle potentially modified targets during preprocessing. An easy solution would be to decode and write back the targets after processing the data item to the dataframe and exporting it alongside the data in the `features_subdir`. Right now, none of our transforms modifies the targets so i left it out here.

Also removed the `RGBAToGrayscale` transform in favor of `RGBToGrayscale` just handling both.


autrainer/serving/serving.py
```diff
@@ -334,20 +336,21 @@ def _collect_files(self, directory: str, extension: str, recursive: bool):
...
         return glob.glob(os.path.join(directory, pattern), recursive=True)
 
-    def _preprocess_file(self, x: torch.Tensor) -> torch.Tensor:
-        x = self._pad_audio(x)
-        x = self.preprocess_pipeline(x, 0)
-        x = self.inference_transform(x, 0)
-        x = x.unsqueeze(0).to(self._device)
-        return x
+    def _preprocess_file(self, x: torch.Tensor) -> DataBatch:
+        item = DataItem(self._pad_audio(x), 0, 0)
```
ATriantafyllopoulos, line 340: I wonder what will happen if the user wants to write a custom `DataItem` struct for his pipeline. It's not that we supported it right now, anyway, but do you have any thoughts?

E.g. sending in the target image + anchors for object detection (or target audio + anchors for SED, or whatever)

ramppdev: Good point, I would say anchors and such would usually be calculated during training and hosted by the model directly.

I guess this PR should only represent the general migration to structs without handling any downstream edge cases that might come up down the line?

autrainer/augmentations/spectrogram_augmentations.py
```diff
@@ -85,19 +88,22 @@ def offset_generator_seed(self, offset: int) -> None:
...
         t = torch.randint(
-            -self.time_steps, self.time_steps, (1,), generator=self._generator
+            -self.time_steps,
+            self.time_steps,
+            (1,),
+            generator=self._generator,
         ).item()
         if self.axis == 1:
-            t1, t2 = x[:, :, :-t].clone(), x[:, :, -t:].clone()
+            t1, t2 = item.features[:, :, :-t], item.features[:, :, -t:]
```
ATriantafyllopoulos, line 102: Is the clone no longer needed? Not sure why we might have needed it in the first place..

autrainer/transforms/specific_transforms.py
```diff
@@ -110,17 +113,19 @@ def __init__(
...
-        data -= data.min()
-        data /= data.max()
-        data = np.uint8(self._cmap(data)[..., :3] * 255)
-        im = Image.fromarray(data)
-        im = im.transpose(Image.FLIP_TOP_BOTTOM)
-        im = im.resize((self.width, self.height))
-        return torch.from_numpy(np.array(im)).transpose(0, 2).to(torch.uint8)
+    def __call__(self, item: AbstractDataItem) -> AbstractDataItem:
+        if item.features.dtype == torch.uint8:
+            item.features = item.features / 255
```
ATriantafyllopoulos, line 118: This was a bug then?

ATriantafyllopoulos (approved):

autrainer/augmentations/spectrogram_augmentations.py
ramppdev: Yeah i was not sure why it was there so i removed it since there are no in-place ops.

PR #148, opened 2025-05-05
Adding DCASE2017-task3 files (TUT-SED)

wagnerp4:
This branch contains the files for the second sound event detection (SED) dataset from TUT, which uses strong labels. The dataset contains multiple folds, which can be specified in the configuration file.


autrainer/datasets/abstract_dataset.py
ramppdev, line 1: Looking at the changes made in the abstract dataset, i thing this branch is not up to date with the base branch as they were probably not intended (also the reason for failing tests). Probably best to rebase real quick to align them again.

PR #149, opened 2025-05-07
Add TUT dataset files

wagnerp4:
This branch contains the files for the TUT Sound Event Detection Dataset, DCASE2017 Task 3.

### DIFF
--- a/autrainer-configurations/dataset/DCASE2017Task3-32k.yaml
+++ b/autrainer-configurations/dataset/DCASE2017Task3-32k.yaml
@@ -1,0 +1,21 @@
+id: DCASE2017Task3-32k
+_target_: autrainer.datasets.DCASE2017Task3
+
+fold: 1
+
+path: data/DCASE2017
+features_subdir: log_mel_32k
+index_column: filename
+target_column: event_label
+file_type: npy
+file_handler: autrainer.datasets.utils.NumpyFileHandler
+
+criterion: torch.nn.BCEWithLogitsLoss
+metrics:
+  - autrainer.metrics.SegmentBasedF1
+  - autrainer.metrics.EventBasedF1
+  - autrainer.metrics.SegmentBasedErrorRate
+tracking_metric: autrainer.metrics.SegmentBasedF1
+
+transform:
+  type: grayscale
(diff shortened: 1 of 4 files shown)


ramppdev:
the `sed_eval` package has to be added to `pyproject.toml`

ramppdev:
> > the `sed_eval` package has to be added to `pyproject.toml`
> 
> 
> 
> I am looking into this, poetry has some python version issues and I have some trouble with creating the lock file. I can address it on the metric pr (#137) as I hope to resolve that soon

What exactly? Version mismatches?

wagnerp4:
> > > the `sed_eval` package has to be added to `pyproject.toml`
> > 
> > 
> > I am looking into this, poetry has some python version issues and I have some trouble with creating the lock file. I can address it on the metric pr (#137) as I hope to resolve that soon
> 
> What exactly? Version mismatches?

When using `poetry==2.0.0` or `1.7.1` the `poetry lock` (or `update`) command gets stuck for me for a long time (over hours without completion) for different python versions on different operating systems. 
I managed to fix it on `poetry==2.0.0` for example by upgrading to `Python=^3.10` in the pyproject.toml - then it resolved in ~30 seconds (but obviously limits the range of version matches). I was trying python 3.9.8, the closest version to 3.10.0 while stickign to `Python=^3.9` but it still gets stuck.

With poetry==1.8.3 (the version of the previous .lock file) the `poetry lock` (or `update`) worked immediately, albeit slower ~100 seconds (using Python 3.10.0). 

Currently there are still some tests failing after the `sed_eval` dependency issue. I am looking into that would then try to update `feature/sed-update-metric` with the new .lock file and the fixes.

ramppdev:
Hm yeah I understand, welcome to dependency hell :D 
I guess at some point we should migrate to Python 3.10 as end of life for 3.9 is this year.
Migrating to Poetry 2.0 does also make sense. Would you create an issue for the migration?

wagnerp4:
> Hm yeah I understand, welcome to dependency hell :D I guess at some point we should migrate to Python 3.10 as end of life for 3.9 is this year. Migrating to Poetry 2.0 does also make sense. Would you create an issue for the migration?

Haha, yeah I've been stuck in it :D
I think a migration makes sense too, I will look into it and create and issue in the coming days.

PR #158, opened 2025-07-29
Add warning in _create_windows for sliding window length exceeding au…

edoromanodev:
- Modified `_create_windows` to return window length, stride length,
  and list of window start indices.
- Added a warning if the sliding window length is greater than the audio
  length during inference.
- In such cases, force at least one window starting at 0 to avoid silent failures.
- Improves user experience by providing clear feedback on windowing issues.


ramppdev:
Closing in favor of #160

PR #159, opened 2025-07-29
Added check and warning for audio clips shorter than 10s sliding wind…

revathim007:
This update adds a validation step in the dataset loader to identify audio clips shorter than the required 10-second sliding window duration. If such clips are found, a clear warning is logged to help users understand potential issues during inference. This improves the user experience by making the cause of sliding window errors more transparent.


ramppdev:
Could it be that the base branch is not up to date?

ramppdev:
closing as stale

PR #166, opened 2025-08-04
Skip Optimizer Weight Decay for Biases and Normalization Layers

ramppdev:
In line with current literature, this splits the model parameters into two groups by default when `weight_decay` is specified:
* Parameters that should have weight decay applied (e.g., weights of convolutional layers, linear layers).
* Parameters that should not have weight decay applied (e.g., biases, normalization layers).

Can be disabled by passing `apply_weight_decay_to_all: true` in the optimizer config.

### DIFF
--- a/docs/source/modules/optimizers.rst
+++ b/docs/source/modules/optimizers.rst
@@ -12,7 +12,8 @@ Any :ref:`torch optimizer <torch_optimizers>` or :ref:`custom optimizer <custom_
 .. note::
 
    Optimizer :attr:`weight_decay` is not applied to biases and normalization layers by default.
-   To change this behaviour, set :attr:`apply_weight_decay_to_all` to ``True`` in the optimizer configuration.
+   To change this behaviour, set :attr:`apply_weight_decay_to_bias` and :attr:`apply_weight_decay_to_norm`
+   to ``true`` in the optimizer configuration.
 
 
 .. _torch_optimizers:


ATriantafyllopoulos:
Thanks. I admit I was not aware of the problem. I myself cannot think of any argument either in favor or against combining batch/layer norm with L2-norm. The "literature" I find (mostly [blog posts](https://loribeehive.github.io/lori-blog/L2/) and [pytorch forums](https://discuss.pytorch.org/t/weight-decay-in-the-optimizers-is-a-bad-idea-especially-with-batchnorm/16994/15)) is contradictory.

Can you please point to some relevant literature?

---
In general, I am okay with adding the option, I just might keep the default to applying weight norm on all parameters for consistency with pytorch..

ramppdev:
I have not found a paper that explicitly mentions this, but many implementations (especially in the CV space always) exclude biases and all norm layers), the idea is the following:
- bias: does not contribute significantly to overfitting, kinda unnecessary and can even can hinder convergence if decayed (pytorch docs skip biases)
- normalization: i.e., for batch norm, both scale (alpha) and shift (beta) are look like normal trainable parameters to the optimizer, decaying these effectively counteracts their purpose

Here would be a practical example (Ultralytics YOLO optimizer example):
https://github.com/ultralytics/ultralytics/blob/7dd9ec6c80dc0514e067566adca6d51eb9b2774d/ultralytics/engine/trainer.py#L813

ramppdev:
> In general, I am okay with adding the option, I just might keep the default to applying weight norm on all parameters for consistency with pytorch..

I would actually keep it as is, generally most CV frameworks apply this by default, and never tell the user about it (we have it in the docs).



> The "literature" I find (mostly [blog posts](https://loribeehive.github.io/lori-blog/L2/) and [pytorch forums](https://discuss.pytorch.org/t/weight-decay-in-the-optimizers-is-a-bad-idea-especially-with-batchnorm/16994/15)) is contradictory.

I dont really get why decaying e.g., the scale for batch norm is good for some networks. We have `y = scale * x + shift` in a bn layer. Initially scale = 1 and shift = 0. When we decay the scale we just force all activations to collapse towards zero and hinder the bn layer to amplify or do whatever with the input.

ramppdev:
> > When we decay the scale we just force all activations to collapse towards zero and hinder the bn layer to amplify or do whatever with the input.
> 
> Don't we do the same if we force all filters of the convolution towards zero?
> 
> As far as I see, it is simply that weight decay "cancels out" the effects of batch norm.
> 
> But, in general, I am fine with adding this since it's clearly identified in the docs, so go ahead and merge :)

Yes, good point, weight decay applied to any parameter ultimately pulls it toward zero, whether its a conv weight or a batch norm scale / weight. So in a broad sense, yes, decaying conv filters also suppresses their effect on the output.
But i would distinguish:
- conv weights are the main source of representation power; decaying them acts as a capacity control, encouraging simpler filters -> hopefully also encourage generalization
- bn scale exists only to rescale normalized activation -> if we decay towards zero, we are not regularizing the networks capacity, we just mute the activations after they have already been normalized and counter the the BN mechanism, which is there to enable flexible rescaling after normalization

This is also the empirical reasoning all other CV frameworks provide, i would say this holds 1:1 for audio as well, so will merge but allow the user to pick whether biases and / or normalization layers should be excluded from weight decay. That way we should be as flexible as possible.

PR #167, opened 2025-08-04
Add Wildcard to Sphinx Directive Extension

ramppdev:
This adds a `*` wildcard for the `ConfigurationsDirective` sphinx extension, allowing to list all configurations present in a given directory, e.g.,
```rst
.. configurations::
   :subdir: dataset
   :configs: *
   :headline:
```
to list all datasets.


ATriantafyllopoulos:
Thanks. Do you think it makes sense to additionally collect all configs for all data/models/etc. under a dedicated header?

PR #172, opened 2025-08-05
Expose Trainer with API

ATriantafyllopoulos:
This draft(!) PR aims to expose a new `Trainer` class to the user so that they can use `autrainer` through a "proper" Python API. Currently, I am just trying to make minimal changes to `ModularTaskTrainer` and expose all attributes for initialization via `__init__` (not fully there yet).

Before I continue, I would like a second pair of eyes on it..

I also add `api_test.py` to play around with the new API. However, I am unable to reproduce the results of `autrainer train` with the default `config.yaml` right now, and different runs of `api_test.py` produce different results, which means I have messed up something in the seed setting..

### DIFF
--- a/api_test.py
+++ b/api_test.py
@@ -1,0 +1,36 @@
+import torch
+
+from autrainer.criterions import BalancedCrossEntropyLoss
+from autrainer.datasets import ToyDataset
+from autrainer.models import FFNN
+from autrainer.training import Trainer
+
+
+if __name__ == "__main__":
+    dataset = ToyDataset(
+        task="classification",
+        size=1000,
+        num_targets=10,
+        feature_shape=64,
+        dev_split=0.2,
+        test_split=0.2,
+        seed=1,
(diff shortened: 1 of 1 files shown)


autrainer/training/training.py
```diff
@@ -874,3 +878,253 @@ def cfg(self) -> DictConfig:
...
+        self.criterion = criterion
+        self.reduction = "none"
+        if hasattr(self.criterion, "setup"):
+            self.criterion.setup(self.data)
+        self.criterion.to(self.DEVICE)
+
+        # ? If no training seed provided, use data seed
+        if training_seed is None:
+            training_seed = self.data.seed
+        set_seed(training_seed)
```
ramppdev, line 944: Setting the seed this late may have an effect on stuff like the model initialization maybe? In the `ModularTaskTrainer` i call `set_seed` as early as possible.

PR #173, opened 2025-08-07
Migrate Dependency Mgmt to uv and drop Python 3.9

ramppdev:
This migrates the dependency management to use `uv` and drops Python 3.9 as EOL is in a couple of months. Also replaces Poetry in the pipelines.

The big difference is in how we manage torch, torchvision, and torchaudio. Similar to Poetry, uv absolutely hates the version extensions (e.g., `2.8.0+cu129`) which specify the cuda version on windows. The only workaround is to first remove and then add back torch etc. as dependecy overrides. This has the advantage that on linux and macos you dont have to install torch manually for dev setup (see contributing section in the docs).
Technically this is also true for pip installation, ive rewritten the docs accordingly, is that fine?

I took the chance to upgrade to torch 2.8, then speechbrain annoyed me with some random prints (not warnings), so now its lazily imported if thats fine.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,9 +20,6 @@ for computer audition tasks using YAML configuration files and the command line.
 Installation
 ------------
 
-To install `autrainer`, first ensure that PyTorch (along with torchvision and torchaudio) version 2.0 or higher is installed.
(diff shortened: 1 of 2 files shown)


ramppdev:
Ah the CI names have changed now, we need to correct this before merging. This will affect all open PRs as well though.

PR #176, opened 2025-09-05
Add `autrainer eval` CLI Command

ramppdev:


ramppdev:
> `dataset@training`

I get the idea, but if we allow sth like `dataset@training` then the user again has to specify all the above config parameters to identify the correct run, so i would either choose the simple or the full option.

> you mean to able to save the evaluation results in a different location? Maybe also more as an optional parameter?

Yes, as optional params so you can run evaluation from anywhere (if we keep the full config version), the slimmed down version you like more would not need this at all.

TzuLingChing:
I think the simpler version with loading the checkpoints is more universal, as I don’t need to rely on previous training runs to do the evaluation. Also allows me to import models from external sources directly.
Otherwise it was all functionality we need for cross data validation.
Thanks for the fast response!

ramppdev:
Should we release this @millinma @ATriantafyllopoulos?

PR #190, opened 2025-11-11
Fix Transform Setup Order

### LINKED ISSUE #189: Standardizer not correctly initialized
@ramppdev it appears that the `Standardizer` is not working in the new version (with the global `setup`).

To reproduce the issue, just change the default `ToyTabular-C.yaml` to:

```yaml
id: ToyTabular-C
_target_: autrainer.datasets.ToyDataset

task: classification
size: 1000
num_targets: 10
feature_shape: 64
dev_split: 0.2
test_split: 0.2

criterion: autrainer.criterions.BalancedCrossEntropyLoss
metrics:
  - autrainer.metrics.Accuracy
  - autrainer.metrics.UAR
  - autrainer.metrics.F1
tracking_metric: autrainer.metrics.Accuracy

transform:
  type: tabular
  base:
    - autrainer.transforms.Standardizer
```

The problem is caused because

a) Global transforms are first initialized when the [create_train_loader](https://github.com/autrainer/autrainer/blob/3ebbe35ab91e354d0951aadd1a8bbd64eb567308/autrainer/datasets/abstract_dataset.py#L273) and similar attributes are called

b) These are first instantiated [here](https://github.com/autrainer/autrainer/blob/3ebbe35ab91e354d0951aadd1a8bbd64eb567308/autrainer/training/training.py#L221)

c) However, we're already calling the transforms [here](https://github.com/autrainer/autrainer/blob/3ebbe35ab91e354d0951aadd1a8bbd64eb567308/autrainer/training/training.py#L165) for `bookkeeping`

@ramppdev do you agree that the simplest fix is to move the creation of the loaders further up in the `ModularTaskTrainer.__init__`?

Issue first reported and debugged by @moOritzl

ramppdev:
Closes #189 by calling transform `setup` at the end of `AbstractDataset` such that the initialization does not care about the loaders being initialized.

Also removes some misc unused loader default kwargs.

### DIFF
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -450,12 +450,8 @@ class TestStandardizer:
             t(DataItem(torch.randn(3, 32, 32), 0, 0))
 
     def test_invalid_setup(self) -> None:
-        dataset = self._mock_dataset(
-            [4, 3, 32, 32],
-            SmartCompose([Standardizer()]),
-        )
         with pytest.raises(ValueError, match="Unsupported data dimensions"):
-            dataset.train_transform.setup(dataset)
+            self._mock_dataset([4, 3, 32, 32], SmartCompose([Standardizer()]))
 
     def test_setup(self) -> None:
         dataset = self._mock_dataset(


ramppdev:
Forgot to update the tests, any setup error will now obviously be risen during dataset init.

